PR #8, opened 2022-10-26
[CLEANING] Remove unused commands.rs module

JasterV:
Just removing a module that is not being used on the server


darioalessandro (approved):
lgtm!!

PR #12, opened 2022-11-05
[Cleaning] Refactor Host and Attendant

JasterV:
I noticed that the whole Host and Attendant streaming logic was inside the update function and I wanted to structure the code on a way that is easier to understand and work with :rocket:

Some notes about these changes:

* Separate Host audio & video initialization/loop into separate functions
* Create private functions to separate message handling from the update function
* Removed the `connected` boolean in the Attendant and Host and instead added a `state` value that will allow to know more precisely in which state the actor is in.

My idea is to decouple the streaming & internal state changes from the yew component, this could be improved by creating a separate service that does not implement `Component` and has all the business logic in it.
Then the `Component` model would just contain an instance of this new service.

I haven't tested yet this locally but so far it is compiling so I'm quite confident it will work


yew-ui/src/components/host.rs
```diff
@@ -50,189 +55,206 @@ pub struct MeetingProps {
...
+                self.initialize(ctx);
                 true
             }
         }
     }
 
     fn view(&self, ctx: &Context<Self>) -> Html {
         // Initialize component on first render.
-        if !self.initialized {
+        if let State::Created = self.state {
```
darioalessandro, line 257: @JasterV We found an easier way to do this using the yew rendered method: 

https://github.com/security-union/rust-zoom/blob/dde8d10d568e1b382cee8f61f749cab98b2414c1/yew-ui/src/components/host.rs#L62-L66

so we removed this state flag all together

yew-ui/src/components/host.rs
```diff
@@ -50,189 +55,206 @@ pub struct MeetingProps {
...
+                    on_audio.emit(audio_frame);
+                }
+                Err(e) => {
+                    log!("error", e);
+                }
+            }
+        }
+    }
+
+    async fn get_device() -> MediaStream {
```
darioalessandro, line 191: @JasterV this is awesome!!!!!!!

yew-ui/src/components/host.rs
```diff
@@ -50,189 +55,206 @@ pub struct MeetingProps {
...
+                    video_frame.close();
+                }
+                Err(e) => {
+                    log!("error", e);
+                }
+            }
+        }
+    }
+
+    async fn start_audio(
```
darioalessandro, line 153: YEEEESS!

darioalessandro (approved):
Hey @JasterV !!!!

Thank you for cleaning my mess!! 

If you remove the self.state property I think we can merge this 😄 !

yew-ui/src/components/host.rs
JasterV: should I remove the state only on the host or also on the attendants?

JasterV:
@darioalessandro  I've applied the requested changes + decoupled business logic from the Yew Component code
I'm not sure if this adds much value to the project but I find the way I've restructured the code easier to understand and maintain.

Some people could disagree with this since I've splitted everything into so many different files,
But I like the pattern of having a Model, Msg and a Component.
The Model acts as the state of the Component and should handle the business logic 
The component handles messages and calls the model with the proper data, then it renders the view using the model data

I think I might have some vias since I've been the last months working with Elm, but as far as I know Yew is highly inspired by the [Elm architecture](https://guide.elm-lang.org/architecture/)

darioalessandro:
Hey @JasterV I am having trouble compiling the project :(

darioalessandro:
Hey @JasterV I took a stab at merging this but it is too big :( I'll take a look later this week.

JasterV:
> Hey @JasterV I took a stab at merging this but it is too big :( I'll take a look later this week.

No worries! We can go over it together at some point this week, I would love to pair on this :)

JasterV:
I think the best approach is going to be to apply the changes you made on the your last PR 1 by 1 and making sure everything works fine, and I think its going to be faster if we do it together

yew-ui/src/components/host/model.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    async fn start_streaming(
+        email: Box<String>,
+        on_frame: Box<Callback<MediaPacket>>,
+        on_audio: Box<Callback<AudioData>>,
+        destroy: Arc<AtomicBool>,
+    ) {
+        let device = Self::get_device().await;
+        join!(
```
darioalessandro, line 205: I love how clean this is!

darioalessandro (changes requested):
Hey @JasterV I am testing locally and the video stops streaming after a few frames with error:

```write chunk error```

darioalessandro:
ok, sounds good 😄

JasterV:
I don't think I'm going to keep working on this PR, I'll close it

PR #35, opened 2023-05-24
Send heartbeat every 1 seconds when connected

### LINKED ISSUE #34: When video and audio is not streaming connected users do not appear in the call
We need to forward websocket pings to all hosts so they know that those users are connected while muted and not showing video

griffobeid:
Closes #34


protobuf/types/media_packet.proto
```diff
@@ -5,6 +5,7 @@ message MediaPacket {
     VIDEO = 0;
     AUDIO = 1;
     SCREEN = 2;
+    HEARTBEAT = 3;
```
griffobeid, line 8: Should add heartbeat_metadata to show if the peer is muted or video is disabled

darioalessandro (approved):
lgtm

PR #41, opened 2023-05-27
Add device selector dropdown

darioalessandro:
@griffobeid is horrible but it works!!

Screen Shot 2023-05-27 at 7 39 11 PM


griffobeid:
Just needs some css 😂

PR #54, opened 2023-06-11
feature: bot dotenv

leon3s:
It's a pain on windows to add env variable so i added a dotenv.
Another solution would have been to use `clap` maybe and take command line arguments!
I also changed the code a bit so it's executed in parallel ! It's an equivalent of a `Promise.all` in javascript!

### DIFF
--- a/bot/src/main.rs
+++ b/bot/src/main.rs
@@ -1,5 +1,6 @@
 use futures::SinkExt;
 use futures::StreamExt;
+use futures::stream::FuturesUnordered;
 use rand::Rng;
 use types::protos::media_packet::MediaPacket;
 use types::protos::media_packet::media_packet::MediaType;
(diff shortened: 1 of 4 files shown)


bot/src/main.rs
```diff
@@ -13,21 +14,19 @@ use chrono::Utc;
 
 #[tokio::main]
 async fn main() {
+    dotenv::dotenv().ok();
```
darioalessandro, line 17: makes sense! is there a way to signal this to users in the README?

bot/src/main.rs
```diff
@@ -13,21 +14,19 @@ use chrono::Utc;
...
-    let mut clients = Vec::new();
-    for _ in 0..n_clients {
-        clients.push(create_client(&endpoint, &room, &echo_user, &email_prefix).await);
-    }
-
-    for client in clients {
-        client.await;
-    }
+    (0..n_clients)
+    .into_iter().map(|_| async {
```
darioalessandro, line 26: neat!!!

darioalessandro (approved):
lgtm! imo it is worth throwing a line in the README about this: https://github.com/security-union/zoom-rs/tree/main/bot

leon3s:
I guess i'm allowed to merge when approved ?

darioalessandro:
> I guess i'm allowed to merge when approved ?

Yes, but you should accept the invite to become a collaborator, you should have it in your email inbox.

leon3s:
Yes i accepted it!

darioalessandro:
@leon3s thank you for all the awesome features!!! you rock!!

PR #77, opened 2023-06-21
refactor chat server join room method

darioalessandro:
Prior to this change, if the server received a message that was not a MediaPacket protobuf message, it would crash while attempting to deserialize it.

I also took the time to refactor the handle a bit into multiple methods

### DIFF
--- a/actix-api/src/actors/chat_server.rs
+++ b/actix-api/src/actors/chat_server.rs
@@ -122,7 +122,13 @@ impl Handler<JoinRoom> for ChatServer {
             if msg.subject == format!("room.{}.{}", room_clone, session_clone) {
                 return Ok(());
             }
-            let msg = MediaPacket::parse_from_bytes(&msg.data).unwrap();
+            let msg = match MediaPacket::parse_from_bytes(&msg.data) {
+                Ok(msg) => msg,
+                Err(e) => {
+                    error!("error parsing message: {}", e);
+                    return Ok(());
+                }
+            };
             let msg = Message {
                 nickname: Arc::new(Some(msg.email.clone())),
                 msg: Arc::new(msg),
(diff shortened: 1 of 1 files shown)


actix-api/src/actors/chat_server.rs
```diff
@@ -105,46 +105,86 @@ impl Handler<JoinRoom> for ChatServer {
...
+    session: String,
+) -> impl Fn(nats::Message) -> Result<(), std::io::Error> {
+    move |msg| {
+        if msg.subject == format!("room.{}.{}", room, session) {
+            return Ok(());
+        }
+
+        let media_packet = match MediaPacket::parse_from_bytes(&msg.data) {
+            Ok(media_packet) => media_packet,
+            Err(e) => {
```
darioalessandro, line 174: @griffobeid if we fail to deserialize packet, log the error and move on

leon3s: Why returning an `Ok` instead of an `Err` ?

leon3s (changes requested):

actix-api/src/actors/chat_server.rs
```diff
@@ -105,46 +105,86 @@ impl Handler<JoinRoom> for ChatServer {
...
+                error!("error parsing message: {}", e);
+                return Ok(());
```
leon3s, line 176: Why return an `Ok` instead of an `Err` ?
On line `187` you return an error if the message cannot be sent

leon3s (changes requested):

actix-api/src/actors/chat_server.rs
darioalessandro: You are right this is a bug, let me fix it

darioalessandro:
@leon3s done, what do you think?

leon3s (approved):
Perfect!

PR #83, opened 2023-06-27
Play packets in order and drop packets that arrive out of order

griffobeid:
Closes #80

Issue #74, opened 2023-06-18
Factor underlying client-side engine out of yew app, for use in other clients


ronen:
I'd like to embed a zoom-rs client in an existing web / electron app, written in javascript using a non-yew UI framework.    That would mean decoupling the encoding/decoding code and control mechanisms from the yew UI code, factoring them out to create an "engine" that could be used in other contexts.   (Something like this would be needed also for  https://github.com/security-union/zoom-rs/pull/49#issuecomment-1587789179 which suggests using electron on MacOS.)

I'm imagining another top-level directory `client-engine/` that would hold that code.  It would be set it up so that it would both:
*  define a local crate that could be imported into the `yew-ui` app
*  compile into a wasm npm module that could be imported into javascript apps.

The engine code wouldn't require specific HTML and wouldn't create any HTML.  It would just provide a control API for establishing connections etc,  it would take `id`s of the canvas elements to render into, and it would use callbacks to communicate interesting events back to the app.

My approach would be to first factor out the client engine code but keep it within `yew-ui/src/`-- and once that looks good, snip it out and move it into that other directory and set up the cargo & npm mechanisms.  The factoring could be done incrementally with PRs along the way, allowing reviews/suggestions as it goes and avoiding a massive shock to the code.

What do you think?  I'm happy to take this on.

darioalessandro:
Hey @ronen !! 

I do like this because I am almost sure that by mixing render code and model we are missing on:

1. performance.
2. cleaner architecture (MVVM?)

ronen:
Cool!  I've already started working on this :)

I'll see what I can do about making small PRs -- but I want to be sure it works before taking the code down a path that may not prove fruitful.   Hopefully something coming up soon.

ronen:
> I'll see what I can do about making small PRs -- but I want to be sure it works before taking the code down a path that may not prove fruitful. Hopefully something coming up soon.

Well, even though I don't have the whole thing done yet, seeing #83 go by made me realize you're right, I better PR what I can while I can :)    So #85 is an initial step.

ronen:
#90 was waiting in the wings for #85 to be ready :)

ronen:
Hi guys -- here's another commit waiting in the wings:  

022476c "Factor out web media connection and use into model::connection::Connection"

[...]

PR #85, opened 2023-06-27
Factor out AudioPeerDecoder and VideoPeerDecoder

ronen:
[First step towards #74]
[Also includes the sequencing fixes from #83]

Factored out structs `AudioPeerDecoder` and `VideoPeerDecoder`, into a new submodule `model::decode`, so that the `Attendants` component no longer has any decoding logic in it.        

Also moved `VideoDecoderWithBuffer` and `configure_audio_context` into model::decode, since they are only called by the  decoders.   All towards the goal of making `model::decode` a standalone bit that could be lifted into an eventual `client-engine` cate.

While doing this, DRY'd the code a bit:  `VideoPeerDecoder` gets used for both video and screen share;  and `AudioPeerDecoder` and `VideoPeerDecoder` are both instantiations of a generic `PeerDecoder<...>` that has the common keyframe and sequence mechanism.

And merged in the changes from #83 even though it hasn't yet been merged into `main`, since it looks like that's on its way to being accepted, and the relocation of the decoders in this PR would breaks auto merge.


darioalessandro:
I just started to look at this, thank you for your contribution

darioalessandro:
Hey @ronen I tested audio & video and lgtm.

Screen share seem to be sad, plus I see that the system is logging something on every frame:
Screenshot 2023-07-01 at 1 32 36 AM

ronen:
> Screen share seem to be broken, plus I see that the system is logging something on every frame, this is on the browser that is streaming the screen: Screenshot 2023-07-01 at 1 32 36 AM

I get the same behavior in `main`, so I don't think it's due to this PR...   plus I think the errors are coming from encoding rather than decoding.    

Screen share has always been a bit funky when i've tried it, hard to get it to start.  I haven't filed an issue because I haven't tracked down what's going on in a clear manner.  I thought it could have been related to connecting twice from two tabs in the same browser (which is how I'm doing my development/testing) -- I sometimes seem to have to start sharing in both connections then stop it in one of them for the sharing to appear, but I haven't found anything reliable.

darioalessandro:
Hey @ronen can you please take another look with the latest main? Seems like screen sharing is working consistently for me. Yes you are right the log message is on the producer side, is this 100% unrelated to your PR? (I can't get main to print this)

ronen:
Hi @darioalessandro, I tried on the latest main, but I'll try again with a clean build.   

In the meantime...  are we on different platforms?  I'm on an intel x86_64 Mac, running the latest MacOS (13.4), testing in the latest Chrome (Version 114.0.5735.198 (Official Build) (x86_64).   Doing all my testing on localhost, with two Chrome tabs connecting with two different usernames.

darioalessandro:
arm Mac OS with the same version of Chrome

darioalessandro:
I am looking at this again, I think that the issue is that we do not refresh the screen on a couple of spots, I will push this over the line

darioalessandro (approved):
lgtm!!!! this is awesome, thank you so much for your contribution!!

ronen:
> I am looking at this again, I think that the issue is that we do not re-render the component on a couple of spots, I will push this over the line

Cool, thanks for looking into it more!  Just got back from weekend away was about to dive in :)

ronen:
Hey @darioalessandro, I can verify that d760779 makes screen sharing work better for me.   But I notice that commit didn't make it into `main` -- was that intentional?

darioalessandro:
@ronen I could not figure out how to commit to your mr directly (I could have forked your repo) so I'll do it on main in a sec

PR #90, opened 2023-07-03
Factor out CameraEncoder, MicrophoneEncoder, and ScreenEncoder

ronen:
[Another step towards #74]

Factored out CameraEncoder, MicrophoneEncoder, and ScreenEncoder

The only structural change is that previously there was a single bit `destroy: Arc<AtomicBool>` that would stop all encoding when it was set, but now each Encoder has its own. Previously the `host.destroy()` method would set that bit, but now it calls a `.stop()` method for each encoder.

Also: 
 * Moved the transforms from model/mod.rs into model/encode/transform.rs since they're only used in model/encode.

 * Moved the wrappers from model/mod.rs into model/wrappers.rs just so that mod.rs isn't a mixed bag of submodule declarations and code definition.


griffobeid:
Thanks for the contribution @ronen 

I tested it out locally and it looks good to me! @darioalessandro can you take a look?

griffobeid (approved):

darioalessandro:
Wow !!! this is great!! I love how you moved all the model out of the host!!

looking now

darioalessandro:
I am sorry for taking this long! will review again by eod tomorrow July 9

ronen:
> I am sorry for taking this long! will review again by eod tomorrow July 9

July 9, but luckily you didn't specify a year :)

Just joking around, no rush.

PR #91, opened 2023-07-03
Fix UI stuttering by playing key frames asap

darioalessandro:


yew-ui/src/components/attendants.rs
```diff
@@ -414,34 +419,38 @@ impl Component for AttendantsComponent {
...
                                 break;
                             }
                             Ok(result) => {
                                 let done = Reflect::get(&result, &JsString::from("done"))
                                     .unwrap()
                                     .unchecked_into::<Boolean>();
+
+                                let value = Reflect::get(&result, &JsString::from("value")).unwrap();
+                                if !value.is_undefined() {
+                                    let value: Uint8Array = value.unchecked_into();
```
darioalessandro, line 446: @griffobeid we were not collecting all the bytes properly

darioalessandro:
@griffobeid I need help with the state management after reconnect. 

I see 2 bugs:
1. Upon disconnect, seems like webtransport crashes internally over and over and the connection does not get terminated, so the reconnect logic does not trigger. 

this is likely a bug inside yew-webtransport because it is not firing the closed callback.

PR #96, opened 2023-07-05
Factor out decoder manager

ronen:
[Another step towards #74]

Created a PeerDecoderManager (I'm open to a better name!) that handles keeping track of the currently connected peers, dispatching each packet to the appropriate decoder, etc. factored all that out of Attendants.

It provides some callbacks that are set by Attendants: 
*   `on_peer_added(key)` -- forces an update
*   `on_first_frame((key, media_type))` -- forces an update if the media type is screen, this fixes the hack of d760779b
*   `get_video_canvas_id(key)` called to get DOM ID of the canvas element into which video should be rendered
*   `get_share_canvas_id(key)` called to get DOM ID of the canvas element into which screen share should be rendered

In the future, could add `on_last_frame((key, media_type))` and `on_peer_removed(key))` to handle a peer stopping audio, video or screen share or a peer disconnecting entirely, if/when there's protocol to detect those.

### DIFF
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -24,9 +24,6 @@ use yew_webtransport::webtransport::{WebTransportService, WebTransportStatus, We
 
 use super::device_permissions::request_permissions;
 
-// This is important https://plnkr.co/edit/1yQd8ozGXlV9bwK6?preview
-// https://github.com/WebAudio/web-audio-api-v2/issues/133
-
 #[derive(Debug)]
 pub enum WsAction {
     Connect(bool),
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -1,8 +1,5 @@
-use std::collections::HashMap;
-use std::sync::Arc;
-
 use crate::constants::WEBTRANSPORT_HOST;
-use crate::model::decode::{AudioPeerDecoder, VideoPeerDecoder};
+use crate::model::decode::PeerDecodeManager;
 use crate::model::MediaPacketWrapper;
 use crate::{components::host::Host, constants::ACTIX_WEBSOCKET};
 use gloo::timers::callback::Interval;
@@ -12,7 +9,6 @@ use js_sys::JsString;
 use js_sys::Reflect;
 use js_sys::Uint8Array;
 use protobuf::Message;
-use types::protos::media_packet::media_packet;
 use types::protos::media_packet::media_packet::MediaType;
 use types::protos::media_packet::MediaPacket;
 use wasm_bindgen::JsCast;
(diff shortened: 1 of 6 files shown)


darioalessandro:
@ronen we mentioned in our reddit announcement, I hope that is ok with you: 
https://www.reddit.com/r/rust/comments/14vwxfh/we_just_released_our_rust_webtransport/

Thanks for helping us to shape up a proper architecture!!

ronen:
> @ronen we mentioned in our reddit announcement, I hope that is ok with you: 

That's fine, thanks for the shout out. 

More to come...

yew-ui/src/components/attendants.rs
```diff
@@ -287,53 +288,16 @@ impl Component for AttendantsComponent {
...
-                        Err(e) => {
-                            log!("error decoding packet: {:?}", e);
-                            return false;
-                        }
-                    }
-                    false
-                } else {
-                    self.insert_peer(email.clone(), screen_canvas_id);
-                    true
+                if let Err(e) = self.peer_decode_manager.decode(response) {
```
darioalessandro, line 297: @ronen so clean!!! this is awesome dude!

yew-ui/src/model/decode/peer_decode_manager.rs
```diff
@@ -0,0 +1,120 @@
...
+            screen: VideoPeerDecoder::new(&screen_canvas_id),
+        }
+    }
+
+    /// Result is:
+    ///     Ok(media_type, Some(true)) on first frame,
+    ///     Ok(media_type, Some(false)) on other frames
+    ///     Ok(media_type, None) on failure to render
+    ///     Err(e) problem with the packet
+    fn decode(&mut self, packet: &Arc<MediaPacket>) -> Result<(MediaType, Option<bool>), i32> {
```
darioalessandro, line 30: @ronen I think this is awesome! would you please document what the `Option<bool>` is for please?

I got lost a bit myself but is it a flag to indicate if the decode should trigger a re-render?

yew-ui/src/model/decode/peer_decode_manager.rs
```diff
@@ -0,0 +1,120 @@
...
+        let media_type = packet.media_type.enum_value()?;
+        match media_type {
+            MediaType::VIDEO => Ok((media_type, self.video.decode(packet).ok())),
```
darioalessandro, line 33: is there a way to expose the decode error here? how about: 

```
match media_type {
            MediaType::VIDEO => self.video.decode(packet).map(|_chunk| media_type)
            ...

ronen: Yeah I'd be happy to -- I was maintaining the previous behavior in which decode errors were ignored.

darioalessandro: yes, from time to time the encoder craps the bed I forgot about that, whoever calls this method, in this case the component might choose to restart the encoder or to ignore the error.

yew-ui/src/model/decode/peer_decode_manager.rs
ronen: Follow up:  Added a DecodeStatus (though the media_type didn't fit in it)

darioalessandro:
still looking at this, sorry for the delay!

darioalessandro:
Code looks good, I'll just test tonight to make sure that everything still works

darioalessandro (approved):

PR #101, opened 2023-07-14
Add aes with tests and yew-ui github workflow

griffobeid:


.github/workflows/cargo-test.yaml
```diff
@@ -39,3 +44,31 @@ jobs:
...
+
+      - name: Install wasm-bindgen-cli
+        uses: jetli/wasm-bindgen-action@v0.2.0
+        with:
+          version: 'latest'
+
+      - name: Run cargo test
+        run: |
+          cd yew-ui
+          cargo test
```
darioalessandro, line 74: I think that there's a tool called `wasm-pack` that is used to test browser tests

griffobeid: Yeah it's installed in the step before this with wasm-bindgen

darioalessandro (approved):

PR #111, opened 2023-07-27
allow pinning tiles

darioalessandro:
Jul-27-2023 09-00-10

### DIFF
--- a/yew-ui/static/style.css
+++ b/yew-ui/static/style.css
@@ -46,7 +46,7 @@ canvas {
   position: relative;
 }
 
-.grid-item-selected {
+.grid-item-pinned {
   position: absolute;
   top: 0;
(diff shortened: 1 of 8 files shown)


yew-ui/static/style.css
```diff
@@ -46,6 +46,15 @@ canvas {
   position: relative;
 }
 
+.grid-item-pinned {
```
griffobeid, line 49: You could use tailwind in the rust files instead of this but I'm fine with it either way

griffobeid (approved):

yew-ui/static/style.css
darioalessandro: I agree, I am still learning tailwind, I'll come back and simplify this

PR #112, opened 2023-07-27
Factor out media device query and selection

ronen:
[another step towards #74]

This PR factors the DOM code and related management code out of the `Attendants` and `DeviceSelector` components, bundling them into two APIs:

* `MediaDeviceList`:  Handles querying the list of available media input devices, and supports selection from the list.  Provides callbacks for when the list is loaded and when a selection is made.  Note in particular the default device is automatically selected -- and its callback called -- the devices are loaded (assuming the device list isn't empty).

* `MediaDeviceAccess`: Handles requesting permission to use audio and video input devices.   Provides callbacks for permission granted and denied and a status check.  (The core DOM code for this was admittedly already in a separate file `components/device_permissions.rs`, but that's been moved into `model/media_devices/media_device_access.rs` to be bundled with the callbacks and status maintenance.)


yew-ui/src/model/media_devices/media_device_access.rs
```diff
@@ -0,0 +1,62 @@
+use gloo_utils::window;
+use std::cell::Cell;
+use std::sync::Arc;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::JsFuture;
+use web_sys::MediaStreamConstraints;
+use yew::prelude::Callback;
+
+pub struct MediaDeviceAccess {
+    granted: Arc<Cell<bool>>,
```
darioalessandro, line 10: @ronen what's the advantage of `Arc<Cell>` vs AtomicBool?

darioalessandro:
Hey @ronen I caused lint to fail after attempting to merge main, I tried to fix it but I do not have access to your remote :( I think that all you need to do is to run cargo fmt inside the yew-ui folder

yew-ui/src/model/media_devices/media_device_access.rs
ronen: None that I can think of :)   Switched to `AtomicBool` (2ac3fa9)

yew-ui/src/components/attendants.rs
```diff
@@ -72,43 +71,72 @@ pub struct AttendantsComponentProps {
...
         peer_decode_manager.get_video_canvas_id = Callback::from(|email| email);
         peer_decode_manager.get_screen_canvas_id =
             Callback::from(|email| format!("screen-share-{}", &email));
+        peer_decode_manager
+    }
+
+    fn create_media_device_access(ctx: &Context<Self>) -> MediaDeviceAccess {
+        let mut media_device_access = MediaDeviceAccess::new();
+        media_device_access.on_granted = {
+            let link = ctx.link().clone();
```
darioalessandro, line 112: @ronen seems like this callback is getting called even if you refuse access to the camera.

darioalessandro (dismissed):
Hey @ronen Thank you so much for your contributions mate!

I opened the incognito window and attempted to refuse access to the camera and the message did not show up, after further inspection, it looks like the success callback is being called even when the user refused access.

I found the bug in the `pub fn request(&self)` method and left a comment


https://github.com/security-union/videocall-rs/assets/1176339/6025350e-9d93-41e7-8a31-499162088aaf

yew-ui/src/model/media_devices/media_device_access.rs
```diff
@@ -0,0 +1,62 @@
...
+use yew::prelude::Callback;
+
+pub struct MediaDeviceAccess {
+    granted: Arc<AtomicBool>,
+    pub on_granted: Callback<()>,
+    pub on_denied: Callback<()>,
+}
+
+impl MediaDeviceAccess {
+    pub fn new() -> Self {
```
darioalessandro, line 16: @ronen if you pass on_granted and on_denied to the new contructor, then the callbacks do not have to be public.

yew-ui/src/model/media_devices/media_device_access.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    pub fn is_granted(&self) -> bool {
+        self.granted.load(Ordering::Acquire)
+    }
+
+    pub fn request(&self) {
+        let future = Self::request_permissions();
+        let on_granted = self.on_granted.clone();
+        let on_denied = self.on_granted.clone();
```
darioalessandro, line 31: This should be:

``` let on_denied = self.on_denied.clone();```

yew-ui/src/model/media_devices/media_device_access.rs
ronen: Yes, but then you have an API requiring a bunch of positional parameters of the same type, which IMHO doesn't make for a great extensible API.   I'd say better would be to use the "options struct" pattern such as used elsewhere in the code base (or builder pattern), but I was getting too lazy.   (Guess where I fall on the endless should-Rust-have-named-parameters debate? :)

In any case, I'm thinking that once #74 is done, it'd be worth taking a pass over the entire API for the client-side engine, and come up with something clean & consistent (and not succumb to laziness :)

But if you feel strongly I can change it to something else now.

yew-ui/src/model/media_devices/media_device_access.rs
ronen: d'oh!   embarrassing.    (i had that tested and working earlier and somehow must have "cleaned up" the code and broken it.)

good catch.

yew-ui/src/model/media_devices/media_device_access.rs
darioalessandro: I do not feel strongly about it, fixing the callback on the other comment will do 👍

PR #117, opened 2023-07-29
add github actions to cut api and ui images to docker hub

darioalessandro:
I added base images so that CI is even faster!!


leon3s (approved):
LGTM!

PR #130, opened 2023-08-09
remove nested call in rsa pub key reception

darioalessandro:


yew-ui/src/components/attendants.rs
```diff
@@ -304,57 +321,30 @@ impl Component for AttendantsComponent {
...
+                                ctx.link()
+                                    .send_message(Msg::OnOutboundPacket(PacketWrapper {
+                                        packet_type: PacketType::AES_KEY.into(),
+                                        email: ctx.props().email.clone(),
+                                        data,
+                                        ..Default::default()
+                                    }));
+                            }
+                            Err(e) => {
+                                error!("Failed to send AES_KEY to peer: {}", e.to_string());
```
darioalessandro, line 345: This is naive error handling, at least we need to let the user know that they need to refresh the page

darioalessandro: I created a separate ticket for this

darioalessandro: I agree!!

PR #135, opened 2023-08-11
Async nats

griffobeid:


actix-api/src/actors/chat_server.rs
```diff
@@ -3,48 +3,40 @@ use crate::messages::{
...
+use std::{collections::HashMap, sync::Arc};
+use tokio::task::JoinHandle;
 use tracing::{debug, error, info, trace};
 
-use super::chat_session::{RoomId, SessionId};
+use super::chat_session::SessionId;
 
 pub struct ChatServer {
-    nats_connection: nats::Connection,
+    nats_connection: Arc<async_nats::client::Client>,
```
darioalessandro, line 15: I do not think you need the Arc pointer here, Client is clonable

actix-api/src/actors/chat_server.rs
```diff
@@ -86,15 +78,25 @@ impl Handler<Leave> for ChatServer {
...
-    fn handle(&mut self, msg: ClientMessage, _ctx: &mut Self::Context) -> Self::Result {
+    fn handle(&mut self, msg: ClientMessage, ctx: &mut Self::Context) -> Self::Result {
         let ClientMessage {
             session,
             room,
             msg,
             user: _,
         } = msg;
         trace!("got message in server room {} session {}", room, session);
-        self.send_message(&room, &msg.data, session);
```
darioalessandro, line 97: is send message now dead code?

PR #136, opened 2023-08-17
Clean up some unused variable and unused result warnings

ronen:
Kinda following up from #130...

Switch from .map_err().and_then() to match { }

This eliminates warnings about unused variables/result.  Also I think makes it easier to read.

Also switch from .and_then(... Ok()) to .map() to be a little cleaner.


darioalessandro (approved):
lgtm

PR #137, opened 2023-08-19
Factor out connection, peer management, and packet handling into `VideoCallClient`

ronen:
[Continuing down the path of #74]

This PR create a single `VideoCallClient` interface that internally takes care of the connection, packet handling, encryption, peer management, etc. -- moving all that code out of the `Attendants` and `Host` components.

Part of that cleanup included changing the encoders so that they too use the VideoCallClient in their interfaces


yew-ui/src/model/client/video_call_client.rs
```diff
@@ -0,0 +1,318 @@
...
+
+impl PartialEq for VideoCallClient {
+    fn eq(&self, other: &Self) -> bool {
+        Rc::ptr_eq(&self.inner, &other.inner) && self.options == other.options
+    }
+}
+
+impl VideoCallClient {
+    pub fn new(options: VideoCallClientOptions) -> Self {
+        let inner = Rc::new(RefCell::new(Inner {
```
darioalessandro, line 63: Hey @ronen thanks for your huge contributions like always! 

I understand why we are using the RefCell here, so that we can get mutable access to the Inner object, please consider wrapping Inner with RwLock so that we do not miss the compile-time checks.

ronen: I didn't know about RwLock, but I did try Mutex -- but unfortunately either of those requires that the wrapped object satisfies `Send`.   And unfortunately somewhere deep down in the connection task objects there are members that don't satisfy Send.   I imagine there's gotta be a way to work around that, but nothing came to mind immediately and I decided not to go too far down that rabbit hole.

For RefCell I only use try_borrow and try_borrow_mut and I check the results, so it won't panic.  But if the try_borrow fails it doesn't block, it simply skips the operation it was supposed to do... so still not necessarily completely robust: it may drop an operation if there is another operation in progress in another JS "thread" -- so ultimately I think using the sync locks would be better if somebody.   [There was one case where I used `borrow()` which I think would never have failed, I just pushed another commit to obviate it]

ronen:
But oh darn, I just tested e2ee (again) and it's not working now.   So hold off on committing this, I need to debug.

ronen:
Hey, sorry for the long delay -- got distracted by my day job.

I've pushed a commit that fixed the problem, by replacing "callback hell" with a more straightforward approach for handling new peers.   And did some code cleanup along the way (those went hand-in-hand :)

Then while I was in a code cleaning mood I tidied up a few other things in the client & peer_decoder_manager, in a few other commits.

Should be good to go now!

yew-ui/src/model/client/video_call_client.rs
darioalessandro: Give a chance to review this, I think we can replace RefCell if we change the design pattern a bit, this is a common issue in rust

ronen: > Give a chance to review this, I think we can replace RefCell if we change the design pattern a bit, this is a common issue in rust

That'd be great!

I've been deliberately making incremental changes to the original code.  But that didn't necessarily walk me down the best path for the resulting design of all these modules.   A re-think of the design as a whole might well do better.

darioalessandro (approved):
lgtm!!!!

darioalessandro:
I took a stab at rebasing this myself but it is not trivial!

griffobeid (approved):

ronen:
Hey @darioalessandro, @griffobeid any thoughts about how to handle the UI test suite failure?  

It seems to be due to `wasm-bindgen` versionitis.

...and seems unrelated to my having just merged in main and resolved the conflicts? 🤷🏻

darioalessandro:
I think that @griffobeid is looking into this

griffobeid:
@ronen @darioalessandro Yeah it is a version mismatch.

After doing 

```sh
cargo install -f wasm-bindgen-cli --version 0.2.87
```

Tests are passing just fine, don't worry about it with your changes I can clean this up

PR #138, opened 2023-08-31
Enable safari support dario

darioalessandro:


darioalessandro:
I did not forget about this!! this is next in my priorities!

PR #146, opened 2023-11-10
Clean up scripts for local development

ronen:
I've cleaned up `./start_dev.sh` so it works out of the box (for me anyway :) -- default values for environment variables, etc. -- to make it easy to develop locally without going through docker.    And I added some checks for a few prerequisites, though there are probably more I didn't think of.

I also changed `./launch-chrome.sh` so that it doesn't generate a new certificate each time, since that was forcing recompilation.

I've updated the README to include instructions for them.

### DIFF
--- a/launch_chrome.sh
+++ b/launch_chrome.sh
@@ -3,24 +3,22 @@
 set -e
 
 SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+CERTSPATH="$SCRIPTPATH/actix-api/certs"
 
-# generate certificate
+if ! [ -f "$CERTSPATH/localhost.der" ] ; then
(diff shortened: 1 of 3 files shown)


launch_chrome.sh
```diff
@@ -3,24 +3,22 @@
 set -e
 
 SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+CERTSPATH="$SCRIPTPATH/actix-api/certs"
 
-# generate certificate
+if ! [ -f "$CERTSPATH/localhost.der" ] ; then
```
darioalessandro, line 8: Neat!!

darioalessandro (approved):
lgtm!! I know that there are many devs out there that prefer bare metal!

PR #155, opened 2024-05-20
Added new Dockerfile, updated Makefile, Updated README.md

JoeAt0none:
As discussed i created a new Dockerfile with multiple build stages and descriptions how to create a multi-arch build. Updated the makefile for the build step using the new dockerfile, updated the README.md with the buildx requirement.

Dockerfile bumps dependency from rust:1.72-slim to rust:1.78-slim


JoeAt0none:
I will redact it for now, i found an issue with this.

PR #187, opened 2025-02-28
Add cli webtransport client

darioalessandro:


webtransport-cli/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+        // Spawn a new task to handle each stream
+            tokio::spawn(async move {
+                // Read stream to the end
+                let data = stream.read_to_end(usize::MAX).await.unwrap();
+                process_packet(&data);  
+            });
+        }
+    });
+
+    let result = tokio::spawn(async move {
```
darioalessandro, line 36: add this and you'll get all the frames 😄

darioalessandro:
just use https://crates.io/crates/webtranscat

PR #208, opened 2025-03-18
Diagnostics P2

darioalessandro:
Nice diagnostics!!

Screenshot 2025-03-24 at 03 08 49


darioalessandro:
** We still need to hook up the SCREEN bitrate on screen!!

PR #209, opened 2025-03-21
Why Makefile, Just Rest in Rust

Walker-00:
Fixed the double end quotes in Makefile and also added a Justfile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ clippy-fix:
 
 fmt:
 		docker compose -f docker/docker-compose.yaml run yew-ui bash -c "cd /app/yew-ui && cargo fmt"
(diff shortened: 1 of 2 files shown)


darioalessandro:
Hey @Walker-00 thanks for your contribution ;) I'll checkout Just, seems like it is pretty popular ;) 

 why Makefile? because that is the standard in open source since the 70's and it does the thing 😄 .

Walker-00:
I'm just wandering that it would also be great to use rust based tool like justfile for in this case ;)
Don't mind me, maybe I'm just being a super rust fan boi.

darioalessandro:
That is a great reason! Makefile are due for a change!!

darioalessandro (approved):

PR #231, opened 2025-03-27
fix screenshare and reduce gap of initial video frame

darioalessandro:


videocall-client/src/encode/screen_encoder.rs
```diff
@@ -215,19 +248,26 @@ impl ScreenEncoder {
...
+            };
 
             let screen_reader = screen_processor
                 .readable()
                 .get_reader()
                 .unchecked_into::<ReadableStreamDefaultReader>();
 
             let mut screen_frame_counter = 0;
-
+            log::info!("Screen reader: {:?}", screen_reader);
```
darioalessandro, line 270: remove this

PR #238, opened 2025-03-29
Show correct aspect ratio, disable screen while not streaming

darioalessandro:


videocall-client/src/constants.rs
```diff
@@ -13,10 +13,6 @@ pub static VIDEO_CODEC: &str = "vp09.00.10.08"; // profile 0,level 1.0, bit dept
 pub const AUDIO_CHANNELS: u32 = 1u32;
 pub const AUDIO_SAMPLE_RATE: u32 = 48000u32;
 
-pub const VIDEO_HEIGHT: i32 = 720i32;
```
darioalessandro, line 16: this is a big win!!

videocall-client/src/decode/peer_decoder.rs
```diff
@@ -94,12 +91,22 @@ impl VideoPeerDecoder {
...
+            let height = video_chunk.display_height() as u32;
+
+            // Set canvas dimensions to match video frame
             render_canvas.set_width(width);
             render_canvas.set_height(height);
-            if let Err(e) = ctx.draw_image_with_html_image_element(&video_chunk, 0.0, 0.0) {
-                error!("error {:?}", e);
+
+            // Clear the canvas and draw the frame
+            ctx.clear_rect(0.0, 0.0, width as f64, height as f64);
```
darioalessandro, line 104: do you really need to clear?

PR #248, opened 2025-04-01
Add swift bindings

darioalessandro:


videocall-client/src/decode/peer_decode_manager.rs
```diff
@@ -52,7 +52,7 @@ impl Display for PeerDecodeError {
 }
 
 pub struct Peer {
-    pub audio: AudioPeerDecoder,
```
darioalessandro, line 55: revert changes to this file!!

videocall-ios/VideoCallKit/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
darioalessandro, line 1: I do not think we need this crap

PR #266, opened 2025-05-21
Test media track stream processor add wasm opus encoder

darioalessandro:


.github/workflows/upload-ui-docker-hub.yaml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - main
+      - test-MediaTrackStreamProcessor-add-wasm-opus-encoder
```
darioalessandro, line 10: revert this

PR #275, opened 2025-06-04
Feature/gh 274 enhance mobile grid layout

darioalessandro:


darioalessandro:
Hey @majorrawdawg 👍 

great progress here.

1. I really liked the new grid system, works great.

2. I think it is a great idea to allow adding and removing peers for dev purposes, please keep that code, just put it behind a `#[cfg(env_var_set = "ENABLE_FAKE_PEERS")]`. 

3. Regarding the list vs grid mode, it does not seem to do anything on desktop so I would hide it. 

4. On mobile it works great, please default to grid mode = true

5. on mobile, the user thumbnail, gets right behind the buttons, would you please move it up a bit so that it is not occluded?
Screenshot 2025-06-04 at 00 46 57

6. I can't find the dropdowns to change mic/video camera, are those gone/hidden? 
7. 
Screenshot 2025-06-04 at 00 40 31

darioalessandro (commented):
Hey @majorrawdawg 👍 

great progress here.

1. I really liked the new grid system, works great.

2. I think it is a great idea to allow adding and removing peers for dev purposes, please keep that code, just put it behind a `#[cfg(env_var_set = "ENABLE_FAKE_PEERS")]`. 

3. Regarding the list vs grid mode, it does not seem to do anything on desktop so I would hide it. 

4. On mobile it works great, please default to grid mode = true

5. on mobile, the user thumbnail, gets right behind the buttons, would you please move it up a bit so that it is not occluded?
Screenshot 2025-06-04 at 00 46 57

6. I can't find the dropdowns to change mic/video camera, are those gone/hidden? 
7. 
Screenshot 2025-06-04 at 00 40 31

majorrawdawg:
For the grid, yeah, it was intended as a permanent mode on mobile, and only some of its traits on desktop (scaling the window will fill with more columns).

But yeah, I'll hide the button, as it is only for test mode.

Regarding the disappearing selectors, whoops!  That's not what I was supposed to do
¯\_(ツ)_/¯ 


I'll fix that.

darioalessandro:
How's it going dude?

PR #304, opened 2025-07-03
Simple clean target

mgalexander:
Plenty of improvement possible here, but this is what I threw together to ensure I could start from known state. Use in good health and I'm happy to walk forward improvements (logging and maybe doing an inventory of the containers, images, volumes instead of just throwing out what I am currently seeing feel like good ideas.)


Makefile
```diff
@@ -19,3 +19,28 @@ fmt:
 
 check: 
 		docker compose -f docker/docker-compose.yaml run websocket-api bash -c "cd /app && cargo clippy --all  -- --deny warnings && cargo fmt --check"
+
+clean:
+		docker compose -f docker/docker-compose.yaml down --remove-orphans
```
darioalessandro: this is pretty good, I think that 

```
 docker compose -f docker/docker-compose.yaml down \
       --remove-orphans  \
       --volumes         \
       --rmi all
       ```
will delete all volumes and images in the compose file:

```
      --remove-orphans   Remove containers for services not defined in the Compose file
      --rmi string       Remove images used by services. "local" remove only images that
                         don't have a custom tag ("local"|"all")
  -t, --timeout int      Specify a shutdown timeout in seconds
  -v, --volumes          Remove named volumes declared in the "volumes" section of the
                         Compose file and anonymous volumes attached to containers
```

darioalessandro (commented):
this is great, I think that `docker compose down` has an even more succinct way to  remove all images and volumes via --volumes and --rmi

mgalexander:
I didn't even look at the docker compose commands again as I took things out one at a time by hand. I'll verify similar results (and maybe remember why I liked docker more) and make changes. Thanks.

mgalexander:
Letting compose take care of it seems to work on my system. It still leaves the anonymous pg volume left for reasons I don't know. More efficient though than taking it one at a time.

PR #315, opened 2025-07-09
Add safari to neteq

darioalessandro:


neteq/src/web.rs
```diff
@@ -38,6 +38,7 @@ impl WebNetEq {
         let cfg = NetEqConfig {
             sample_rate,
             channels,
+            min_delay_ms: 500,
```
darioalessandro: todo: it is not a good idea to hardcode this!!!

PR #324, opened 2025-07-15
Fix pin icon positioning and visibility on iOS and desktop

iamgabrielsoft:
### Description:
This PR addresses the issue where the pin icon covers the speaker's face on iOS devices by:
1. Moving the pin icon to the top-left corner of the video thumbnail
2. Making the pin icon visible only on hover for desktop users
3. Ensuring the pin icon remains visible on mobile devices

Changes:
- Adjusted pin icon positioning to top-left corner
- Improved visibility handling for different screen sizes
- Enhanced hover behavior for desktop users
- Maintained mobile visibility while avoiding face obstruction

Files changed:
pin-fix.css

Fixes: issue #296

PR #338, opened 2025-07-24
Fix pin icon positioning and visibility on iOS and desktop (#324)

darioalessandro:
* pin hover

* added once_cell

* env file

* default env

* resolve merge conflict

* cargo file

* clipper cleanup

* miss chat_server lint error

* test cleanup

* Update yew-ui/index.html

go on

* consolidating pin.css into global.css

* pkg update

---------

### DIFF
--- a/yew-ui/src/components/audio_diagnostics.rs
+++ b/yew-ui/src/components/audio_diagnostics.rs
@@ -1,0 +1,236 @@
+/*
+ * Copyright 2025 Security Union LLC
+ *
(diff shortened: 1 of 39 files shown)


yew-ui/src/components/audio_diagnostics.rs
```diff
@@ -0,0 +1,236 @@
+/*
```
darioalessandro, line 1: I need to review this in detail

PR #370, opened 2025-08-05
feat(system-specs): host system diagnostics component, Docker fixes

darioalessandro:


yew-ui/Cargo.toml
```diff
@@ -36,7 +36,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 
 [dependencies.web-sys]
-version = "0.3.72"
```
darioalessandro, line 39: do we need to upgrade?

PR #385, opened 2025-08-12
Add packets per sec, and matomo logs to debug system, handle mic errors more gracefully

darioalessandro:


helm/global/us-east/rustlemania-ui/values.yaml
```diff
@@ -15,9 +15,9 @@ rustlemania-ui:
     webTransportEnabled: "true"
     usersAllowedToStream: ""
     serverElectionPeriodMs: 2000
-    audioBitrateKbps: 65
```
darioalessandro, line 18: Reduced quality just to get things to work

Issue #389, opened 2025-08-14
Show self in attendees UI peer list


testradav:
### Problem
- attendants list does not show self as one of the attendants

### Proposed Solution
- always show yourself `username (you)` as the first attendant peer

### Reasoning
- makes sense to see for example a meeting of 3 people, 3 attendants in the list

<img width="804" height="577" alt="Image" src="https://github.com/user-attachments/assets/ff9024e9-cd7c-4c45-8ba5-a9ab638da515" />

darioalessandro:
Yes!! @testradav this is a great point!!! and low hanging fruit!!

testradav:
> Yes!! [@testradav](https://github.com/testradav) this is a great point!!! and low hanging fruit!!

haha, thanks.  getting the hang of RUST...  here is the PR https://github.com/security-union/videocall-rs/pull/390

darioalessandro:
This is done 😄

PR #390, opened 2025-08-14
Update attendants list with self username (yew-ui)

testradav:
Issue #389 

image

### DIFF
--- a/yew-ui/src/components/peer_list.rs
+++ b/yew-ui/src/components/peer_list.rs
@@ -73,14 +73,13 @@ impl Component for PeerList {
             PeerListMsg::UpdateSearchQuery(input.value())
         });
 
-        // Get username from context and append (SELF)
-        let display_name: String = format!(
-            "{} (You)",
-            ctx.link()
-                .context::<UsernameCtx>(Callback::noop())
-                .and_then(|(state, _handle)| state.as_ref().cloned())
-                .unwrap_or_else(|| "(You)".to_string())
-        );
+        // Get username from context and append (You)
+        let display_name: String = ctx
+            .link()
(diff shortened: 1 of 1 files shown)


yew-ui/src/components/peer_list.rs
```diff
@@ -72,6 +73,15 @@ impl Component for PeerList {
             PeerListMsg::UpdateSearchQuery(input.value())
         });
 
+        // Get username from context and append (SELF)
+        let display_name: String = format!(
```
darioalessandro: this lgtm, might show (You) (You) if we fail to get the username, but it is better than what we have now, (much better)

darioalessandro:
@testradav just run `make fmt` to fix `cargo fmt` and we should gtg

yew-ui/src/components/peer_list.rs
testradav: updated to show just (You) if username is not found

darioalessandro:
Just one extra space! 

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 15s
Diff in /app/yew-ui/src/components/peer_list.rs:103:
                             <ul>
                                 // show self as the first item with actual username
                                 <li><PeerListItem name={display_name.clone()} /></li>
-                                
+
                                 { for filtered_peers.iter().map(|peer|
                                     html!{
                                         <li><PeerListItem name={peer.clone()}/></li>
```

testradav:
> @testradav just run `make fmt` to fix `cargo fmt` and we should gtg

I ran `make fmt` but it did not create/change anything to commit... pardon my RUST rookiness

testradav:
Rebuild

Issue #397, opened 2025-08-16
Received unexpected message from worker causing multiple warnings.


darioalessandro:
Here's an screenshot:
<img width="1693" height="1167" alt="Image" src="https://github.com/user-attachments/assets/8ba9088d-78f1-4651-ba3d-e76403143d33" />

Expected behavior, instead of raising a warning message, the stats shall be relayed to the HEALTH system

darioalessandro:
This is the code that we must change:

https://github.com/security-union/videocall-rs/blob/1fd52e54e53b603b0e44b25ec778d7dd196788e3/videocall-codecs/src/decoder/wasm.rs#L155-L167

The problem is that right now the Wasm decoder does not know how to handle the diagnostics message from the worker, all it does right now is to assume that this is a decoded videopacket and sends that upstream to the UI. 

We need to change the contract between the wasm decoder and the worker_decoder, so that wasm decoder handles both statistics and video packets. 



We know that https://github.com/security-union/videocall-rs/blob/main/videocall-codecs/src/bin/worker_decoder.rs is producing both the decoded video frames and the diagnostics messages.

PR #400, opened 2025-08-17
fix(codecs): route worker diag messages to health bus; refs #397

majorrawdawg:
Root cause: worker posts kind="video_stats" objects; main-thread handlers only accepted VideoFrame and warned.

Minimal change in videocall-codecs/src/decoder/wasm.rs onmessage handlers: keep VideoFrame fast path; for non-VideoFrame, if kind==="video_stats" forward to diagnostics; else warn. No hot-path overhead.

### DIFF
--- a/videocall-codecs/src/decoder/wasm.rs
+++ b/videocall-codecs/src/decoder/wasm.rs
@@ -20,7 +20,10 @@
 
 use super::{Decodable, DecodedFrame};
 use crate::frame::FrameBuffer;
+use js_sys::Reflect;
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "wasm")]
+use videocall_diagnostics::{global_sender, metric, now_ms, DiagEvent};
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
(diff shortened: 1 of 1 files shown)


videocall-codecs/src/decoder/wasm.rs
```diff
@@ -260,3 +253,37 @@ impl Drop for WasmDecoder {
         self.worker.terminate();
     }
 }
+
+/// Handle diagnostics objects posted by the worker. Returns true if handled.
+fn handle_worker_diag_message(js_val: &JsValue) -> bool {
+    // Try to deserialize the JavaScript object using serde
+    match serde_wasm_bindgen::from_value::<VideoStatsMessage>(js_val.clone()) {
```
darioalessandro, line 260: @majorrawdawg I ended up using `serde_wasm_bindgen` to parse the json into the stats object to avoid those reflection calls.

majorrawdawg: Ok. I hope my attempt wasn't a total waste.  Glad you knew how to make it better 😊

PR #440, opened 2025-10-09
connection manager events not persisted in yew - use new Global event vector instead

### LINKED ISSUE #436: Broken diagnostics after https://github.com/security-union/videocall-rs/pull/432/files
**Describe the bug**
Connection manager says that the state is unknown

**To Reproduce**
pop open diagnostics and it is all broken XD 

**Additional context**
I broke it in this PR https://github.com/security-union/videocall-rs/pull/432/files

jboyd01:
this is my first real work with Yew (and I'm new to rust) -- I've got no problems if someone has a proper fix using Yew or if you want to take an alternate approach.

fixes https://github.com/security-union/videocall-rs/issues/436

**Issue**
Events collected in the diagnostics component's connection_manager_events state were not persisting between async callbacks in the same component instance, even when the component hadn't remounted. This caused events to appear to disappear between updates, making the diagnostics display unreliable.

**Root Cause**
It appears to me that Yew's state hooks isn't persisting state across async operations that involve yielding control back to the event loop. When an async callback receives a new event, it is operating on a fresh version of the state rather than the accumulated one each time.

**Solution**
Implemented a global state vector using thread-safe primitives to maintain state across all async callbacks. This approach:

1. Creates a static global store for connection manager events
2. Updates this global store first when events are received
3. Ensures component state stays in sync with the global store
4. Handles concurrent access with proper locking

This provides persistence of diagnostic data regardless of when async callbacks are invoked

**Testing**
The fix was verified by confirming that events now accumulate properly across async callback invocations, with the UI correctly displaying the persistent connection manager state.

### DIFF
--- a/yew-ui/src/components/diagnostics.rs
+++ b/yew-ui/src/components/diagnostics.rs
@@ -16,16 +16,47 @@
  * conditions.
  */
 
+//! # Diagnostics Module
+//!
+//! This module provides components for displaying real-time diagnostics information
+//! about WebRTC connections, media streams, and the connection manager's state.
+//!
+//! ## Components
+//!
+//! - `Diagnostics`: The main diagnostics sidebar component that collects and displays all metrics
+//! - `ConnectionManagerDisplay`: Shows connection status, server selection, and election process
+//! - `NetEqStatusDisplay`: Shows network jitter buffer statistics for audio processing
+//!
+//! ## Key Features
+//!
+//! - Real-time connection manager state monitoring
+//! - Server election process visualization
+//! - Audio buffer and jitter metrics with visual charts
+//! - Per-peer diagnostics with filtering
+//! - Media state reporting (video, audio, screen sharing)
+//!
+//! The diagnostics system uses a global state mechanism to persist connection manager events
+//! across asynchronous callbacks, ensuring consistent state tracking even when components
+//! are unmounted and remounted during React/Yew component lifecycle events.
+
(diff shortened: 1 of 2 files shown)


darioalessandro:
I am reviewing this now

darioalessandro:
@jboyd01 your PR inspired me to find a way to store those diagnostics without using global variables and I figured it out!! https://github.com/security-union/videocall-rs/pull/462/files 

thank you for the inspiration

yew-ui/src/components/diagnostics.rs
```diff
@@ -16,16 +16,47 @@
...
+use std::sync::{Arc, RwLock};
+use once_cell::sync::Lazy;
 use videocall_diagnostics::{subscribe, DiagEvent, MetricValue};
 use yew::prelude::*;
+use log::debug;
 
-// Serializable versions of DiagEvent structures (with owned strings instead of &'static str)
+// Global storage for connection manager events to persist across async callbacks
+// This is needed because Yew's state hooks don't preserve state across async callbacks
+static CONNECTION_MANAGER_EVENTS: Lazy<Arc<RwLock<Vec<SerializableDiagEvent>>>> =
```
darioalessandro, line 57: found a way to store this into the component state using `RC<RefCell>

jboyd01:
closed in preference to [PR 462](https://github.com/security-union/videocall-rs/pull/462)

PR #455, opened 2025-10-18
NetEq improvements

adam-kiss-sg:
DRAFT PR FOR DISCUSSION ONLY

Hi!

I sent a couple days on the NetEq code. I mainly plan to use this jitter buffer for TTS audio, so some of the changes might not translate well for cases when there is more noise. Please look at the PR and let me know which parts you like and would consider merging. I can open some smaller proper PRs for the relevant parts if you are interested (although not sure when I will have the time to clean it up). Also, feel free to take anything from this if you want to.

I left in some debug code on purpose, it randomly changes the minimum size of target_level_ms between 30ms and 300ms, so it triggers PreemptiveExpand and Accelerate. As I was testing, it's hardly noticable in the audio.

### Algorithm changes

Updated to use a histogram, and match the webrtc algorithm.

### Opinionated/empirical changes

Updated how minimum and maximum are calculated:
- base_minimum_delay_ms and base_maximum_delay_ms are now exclusively set from NetEq, and represent the pratical limitations based on packet size and
- minimum_delay_ms and maximum_delay_ms can be set from client code. Instead of returning a success/failure, they now return the effective value after setting them
	- rationale:
		- seems more user friendly then returning an error if the max is set too high
		- I didn't like that setting minimum first and then maximum might result in different bounds than setting maximum first and then minimum

Added additional_delay_ms
- this adds the value to the calculated target_level_ms
- I mainly added this to avoid Expand operations as much as possible (min delay wasn't heling if the target_level_ms was already higher than min_delay)

### General changes

Modified all time stretch operations to return a fix length output, instead of consuming a fix length input. They now fill the supplied output buffer, consuming as much of the input as necessary in the process. This makes it possible to keep the size of the returned frames consistent (10ms).

Since both accelerate and preemtive expand works much better with 30ms audio, I changed decode_accelerate and decode_preemptive_expand to produce 30ms output if possible. The first 10ms is returned, and the remaining is stored in a buffer and returned as the next two packets. This buffer is separate from the leftover_samples buffer to avoid double time-streching.

In order to return the extra frames, I added a new TimeStretchBuffer operation.

### Accelerate changes

Optimized the accelerate algorithm. The base logic is the same: remove low energy region and fade over the remaining head and tail. I added a new algorithm that finds the longest region where the average energy is less then a threshold. Instead of removing a fixed sized region, it now removes a region up to max size, where the energy average is below the threshold. 

I tested this with TTS audio, and it works very well. Not sure if it would still work reliably if the audio contained more noise. 

The normal accelerate will do nothing if it doesn't find a low energy region worth removing. The fast accelerate will fall back to the previous method.

#### Opinionated/empirical changes

I changed the low energy threshold form 0.01 to 0.001, this value works better for me. Not sure how this would work with higher background noise.

I also changed overlap length from 5ms to 3ms. 5ms produced an under-water like sound for me, and I heard popping sounds with 2ms. 3ms seems to be a good spot.

### PreemptiveExpand changes

Optimized the algorithm:
- previously: found the best correletion assuming a fix lag (fixed number added samples)
- webrtc: if I understood it correctly: over a range of lag values, finds the best lag value where the correlation of the whole overlapping part of head and tail is the maximum
- my implementation: for each possible lag value, finds the best position for the fade, by maximizing the correlation of only the overlap_length sequence, then returns the best lag value, with the best position for the fade

#### Opinionated/empirical changes

I changed overlap length from 5ms to 3ms, same as for accelerate.

### Expand

Added fade at the start and end of a series of expands. Added two new operations for this: ExpandStart and ExpandEnd. The overlap duration is 5ms, this should probably be shortened as well.

#### Opinionated/empirical changes

Reduced the noise amplitudes from 0.01 to 0.0001. For my use case (TTS audio), the generated noise was way too loud.

## Web

Changed min_delay_ms: 80 to additional_delay_ms: 80, mainly to avoid Expand operations as much as possible.

### DIFF
--- a/neteq/src/delay_manager.rs
+++ b/neteq/src/delay_manager.rs
@@ -19,7 +19,7 @@
 use std::collections::VecDeque;
 use web_time::{Duration, Instant};
 
-use crate::{NetEqError, Result};
+use crate::{histogram::Histogram, Result};
 
 /// Initial target delay when NetEQ starts operation.
 ///
@@ -35,6 +35,9 @@ use crate::{NetEqError, Result};
 /// - Provides safety margin during cold start before statistics stabilize
 const K_START_DELAY_MS: u32 = 80;
 
+const K_DELAY_BUCKETS: usize = 100;
+const K_BUCKET_SIZE_MS: i32 = 20;
+
 /// Configuration for the delay manager
 #[derive(Debug, Clone)]
 pub struct DelayConfig {
@@ -48,10 +51,12 @@ pub struct DelayConfig {
     pub resample_interval_ms: Option<u32>,
     /// Maximum history length in milliseconds
     pub max_history_ms: u32,
-    /// Maximum number of packets in buffer
-    pub max_packets_in_buffer: usize,
     /// Base minimum delay in milliseconds
     pub base_minimum_delay_ms: u32,
+    // Maximum delay in milliseconds
+    pub base_maximum_delay_ms: u32,
+    // Contant additional delay in milliseconds
+    pub additional_delay_ms: u32,
     /// Whether to use reorder optimizer
     pub use_reorder_optimizer: bool,
     /// Forget factor for reorder detection
@@ -63,13 +68,14 @@ pub struct DelayConfig {
 impl Default for DelayConfig {
     fn default() -> Self {
         Self {
-            quantile: 0.95,
-            forget_factor: 0.983,
+            quantile: 0.97,
+            forget_factor: 0.9993,
             start_forget_weight: Some(2.0),
             resample_interval_ms: Some(500),
             max_history_ms: 2000,
-            max_packets_in_buffer: 200,
             base_minimum_delay_ms: 0,
+            base_maximum_delay_ms: 2000,
+            additional_delay_ms: 0,
             use_reorder_optimizer: true,
             reorder_forget_factor: 0.9993,
             ms_per_loss_percent: 20,
@@ -106,15 +112,23 @@ impl RelativeArrivalDelayTracker {
     }
 
     /// Update with a new packet arrival
-    pub fn update(&mut self, timestamp: u32, sample_rate: u32) -> Option<i32> {
-        let arrival_time = Instant::now();
+    pub fn update(&mut self, timestamp: u32, sample_rate: u32, arrival_time: Instant) -> i32 {
+        // Calculate expected time since last packet
+        let expected_iat_ms = if let Some(last_timestamp) = self.last_timestamp {
+            timestamp.saturating_sub(last_timestamp) * 1000 / sample_rate
+        } else {
+            0
+        };
 
-        // Calculate inter-arrival time delay
-        let iat_delay_ms = if let Some(last_time) = self.last_packet_time {
+        // Calculate actual time since last packet
+        let iat_ms = if let Some(last_time) = self.last_packet_time {
             arrival_time.duration_since(last_time).as_millis() as i32
         } else {
             0
         };
+
+        // Calculate delay, positive means packet is late, negative means packet is early
+        let iat_delay_ms = iat_ms - expected_iat_ms as i32;
 
         self.last_packet_time = Some(arrival_time);
 
@@ -155,24 +169,34 @@ impl RelativeArrivalDelayTracker {
             .retain(|delay| now.duration_since(delay.arrival_time) <= max_age);
     }
 
-    fn calculate_relative_packet_arrival_delay(&self) -> Option<i32> {
+    /// Calculates the relative arrival delay of packets in the history.
+    ///
+    /// This effectively computes the accumulated delay of packets relative
+    /// to the packet preceding the history window. If the running sum ever
+    /// goes below zero, it is reset, meaning the reference packet is moved.
+    ///
+    /// Note on behavior:
+    /// - This function is **sensitive to positive tail jitter**:
+    ///     * Positive delays near the end of the history accumulate.
+    ///     * Positive delays near the beginning may be partially cancelled by
+    ///       subsequent negative delays.
+    /// - Negative delays are mostly ignored because the running sum resets
+    ///   to zero whenever it becomes negative.
+    pub fn calculate_relative_packet_arrival_delay(&self) -> i32 {
         if self.delay_history.len() < 2 {
-            return None;
+            return 0;
         }
 
-        // Calculate mean inter-arrival time
-        let total_iat: i32 = self.delay_history.iter().map(|d| d.iat_delay_ms).sum();
-        let mean_iat = total_iat as f64 / self.delay_history.len() as f64;
+        let mut relative_delay: i32 = 0;
 
-        // Calculate relative delay using quantile estimation
-        let mut delays: Vec<i32> = self.delay_history.iter().map(|d| d.iat_delay_ms).collect();
-        delays.sort_unstable();
+        for delay in &self.delay_history {
+            relative_delay += delay.iat_delay_ms;
+            if relative_delay < 0 {
+                relative_delay = 0; // reset if sum goes below zero
+            }
+        }
 
-        let quantile_index = ((delays.len() - 1) as f64 * self.config.quantile) as usize;
-        let quantile_delay = delays[quantile_index] as f64;
-
-        // Return the difference from expected inter-arrival time
-        Some((quantile_delay - mean_iat) as i32)
+        relative_delay
     }
 }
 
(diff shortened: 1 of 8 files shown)


darioalessandro:
Wow! Thank you for your contributions! I'll take a look 👀👀👀👀👀👀

neteq/src/delay_manager.rs
```diff
@@ -155,24 +169,34 @@ impl RelativeArrivalDelayTracker {
...
-
-        // Calculate relative delay using quantile estimation
-        let mut delays: Vec<i32> = self.delay_history.iter().map(|d| d.iat_delay_ms).collect();
-        delays.sort_unstable();
+        let mut relative_delay: i32 = 0;
 
-        let quantile_index = ((delays.len() - 1) as f64 * self.config.quantile) as usize;
-        let quantile_delay = delays[quantile_index] as f64;
+        for delay in &self.delay_history {
+            relative_delay += delay.iat_delay_ms;
```
darioalessandro, line 193: Accumulation instead of percentile 

seems like this is just adding all the delays together and I end up in videocall with a target delay of more than 2000ms instead of the expected [100 - 400]ms.

With 2000ms of packet history, this can accumulate to **2000ms+**, which:
1. Maps to bucket 99 (max bucket: 2000ms / 20ms per bucket)
2. Histogram settles at 97th percentile = bucket 99
3. TARGET = (1 + 99) × 20 = **2000ms**
4. Plus additional_delay_ms (80ms) = **2080ms**

### Impact

- Users experience **2+ seconds of audio latency** instead of ~100ms
- Buffer grows unnecessarily large
- System becomes unusable for real-time communication

### Fix
**1. Calculate jitter correctly (line 115-131):**

```rust
// Calculate expected time since last packet based on RTP timestamps
let expected_iat_ms = if let Some(last_timestamp) = self.last_timestamp {
    timestamp.saturating_sub(last_timestamp) * 1000 / sample_rate
} else {
    0
};

// Calculate actual time since last packet
let actual_iat_ms = if let Some(last_time) = self.last_packet_time {
    arrival_time.duration_since(last_time).as_millis() as i32
} else {
    0
};

// Calculate JITTER (not raw IAT)
let iat_delay_ms = actual_iat_ms - expected_iat_ms as i32;
```

**2. Use percentile instead of accumulation (lines 172-198):**

```rust
pub fn calculate_relative_packet_arrival_delay(&self) -> i32 {
    if self.delay_history.is_empty() {
        return 0;
    }

    // Collect positive jitter values (late arrivals)
    let mut positive_delays: Vec<i32> = self
        .delay_history
        .iter()
        .map(|d| d.iat_delay_ms.max(0))
        .collect();

    if positive_delays.is_empty() {
        return 0;
    }

    positive_delays.sort_unstable();

    // Return 97th percentile of recent jitter
    let index = ((positive_delays.len() - 1) as f64 * self.config.quantile) as usize;
    positive_delays[index]
}
```

neteq/src/time_stretch.rs
```diff
@@ -167,36 +220,85 @@ impl Accelerate {
         best_position
     }
 
-    fn apply_crossfade(&self, fade_out: &[f32], fade_in: &[f32], output: &mut Vec<f32>) {
-        let len = fade_out.len().min(fade_in.len());
+    fn longest_low_energy_region<F>(
```
darioalessandro, line 223: This is a fancy function that I would like to keep 😄

neteq/src/histogram.rs
```diff
@@ -0,0 +1,222 @@
...
+ * at your option.
+ *
+ * Unless you explicitly state otherwise, any contribution intentionally
+ * submitted for inclusion in the work by you, as defined in the Apache-2.0
+ * license, shall be dual licensed as above, without any additional terms or
+ * conditions.
+ */
+
+#[derive(Debug, Clone)]
+pub struct Histogram {
```
darioalessandro, line 20: New Histogram-Based Delay Tracking
Replaces exponential smoothing with Q30 fixed-point histogram
Uses 100 buckets, 20ms per bucket
More robust for bursty jitter
Assessment: Sound approach, matches WebRTC design 

Please do bring this in as a PR 😄 and thank you

neteq/src/delay_manager.rs
```diff
@@ -63,13 +68,14 @@ pub struct DelayConfig {
 impl Default for DelayConfig {
     fn default() -> Self {
         Self {
-            quantile: 0.95,
-            forget_factor: 0.983,
+            quantile: 0.97,
+            forget_factor: 0.9993,
```
darioalessandro, line 72: Why slower adaptation? (forget_factor 0.9993 vs 0.983)

darioalessandro (commented):
Thank you so much for your contributions!!

I left some comments, let me know what you think

neteq/src/time_stretch.rs
```diff
@@ -28,15 +28,10 @@ pub enum TimeStretchResult {
 /// Audio time-stretching processor
 pub trait TimeStretcher {
     /// Process audio samples and return the result
-    fn process(
-        &mut self,
-        input: &[f32],
-        output: &mut Vec<f32>,
-        fast_mode: bool,
-    ) -> TimeStretchResult;
+    fn process(&mut self, input: &[f32], output: &mut [f32], fast_mode: bool) -> TimeStretchResult;
```
darioalessandro, line 31: Better memory management (no allocations during processing)
Stack-based algorithm for finding optimal removal points is clever
More intelligent about when/where to remove samples

please keep these changes

neteq/src/web.rs
```diff
@@ -48,7 +48,7 @@ impl WebNetEq {
         let cfg = NetEqConfig {
             sample_rate: self.sample_rate,
             channels: self.channels,
-            min_delay_ms: 80,
+            additional_delay_ms: 80,
```
darioalessandro, line 51: keep 😄

neteq/src/delay_manager.rs
adam-kiss-sg: This function is actually a copy from the webrtc implementation (https://chromium.googlesource.com/external/webrtc/+/HEAD/modules/audio_coding/neteq/delay_manager.cc#239), and it it was probably the most confusing part for me as well.

The key here is that iat_delay_ms is either a positive or a negative number: if we expect packets to arrive in 20ms intervals, and a packet arrived 30ms after the previous one, than iat_delay_ms will be 10ms. If a packet arrived 8ms after the previous one, it will be -12ms. So regardless of how long the delay_history is, the sum of it's elements will be around 0.

With an example:
Let's say that we have a stream of 20ms packets. We expect these packets to arrive at (ms):
0, 20, 40, 60, 80
But they actually arrive at
2, 21, 45, 63, 87
Calculating actual_iat_ms for each will result in:
0, 19, 24, 18, 24
Then calculating iat_delay_ms: (20 - actual_iat_ms)
0, -1, 4, -2, 4

The calculate_relative_packet_arrival_delay function then calculates a rolling sum, which will be:
0, -1, 3, 1, 5

So, what the simple rolling some actually shows for packet n, is how much packet n was delayed compared to the reference packet (the first packet):

The first (reference) packet in the example is 2 sec later than expected.
2nd packet: -1 ms compared to the reference packet (21-20=1ms, compared to 2ms)
3rd packet: +3 ms compared to the reference packet (45-40=5ms, compared to 2ms)
4th packet: +1 ms
5th packet: +5ms

So, in the simple rolling sum will return the latency of the last packet compared to the latency of the first packet.
But what happens, if the first (or first few) packets in the history were late themselves? That's where the second part comes in:
 ```
             if relative_delay < 0 {
                relative_delay = 0; // reset if sum goes below zero
            }
 ```
 
What this does is, whenever it finds a packet that has lower latency then the reference packet, it will use this packet as the new reference.

So effectively what the function does is:
- find the packet in the history with the lowest latency
- compare the new packet's latency to the found packet

neteq/src/delay_manager.rs
adam-kiss-sg: I copied the values from webrtc, to match the implementation: https://chromium.googlesource.com/external/webrtc/+/HEAD/modules/audio_coding/neteq/delay_manager.cc#42

darioalessandro:
I deployed videocall.rs with all your fixes - the delay manager calculation and I do see performance improvements 😄

darioalessandro:
This is my PR https://github.com/security-union/videocall-rs/pull/466

adam-kiss-sg:
Closing this pr as the changes have been added in https://github.com/security-union/videocall-rs/pull/466

PR #474, opened 2025-11-01
improved neteq expand

adam-kiss-sg:
## What is this PR about?

This PR improves how the NetEq audio buffer handles packet loss concealment - basically, what happens when audio packets don't arrive on time during a call.

## The Main Changes

1. Created a dedicated Expand module (expand.rs)
Previously, the packet loss concealment logic was scattered throughout neteq.rs. Now it's organized into its own clean module with proper structure and tests.
2. Introduced explicit "phases" for packet loss
Instead of using confusing boolean flags (start: bool, end: bool), it now uses a clear enum:
ExpandStart: First moment when packets go missing - crossfade from real audio to generated noise
Expand: Continuing packet loss - just generate quiet noise
ExpandEnd: Packets resume - crossfade from noise back to real audio
3. Much smaller overlap/crossfade regions
Before: Used 50% of the frame for crossfading (way too aggressive)
After: Uses only ~3ms (48 samples at 16kHz) - much more subtle and natural sounding
4. Smarter expand strategy (the key improvement!)
Before: Would stop expanding as soon as ANY packets arrived, even if the buffer was still unhealthy. This caused lots of jarring expand→normal→expand cycles.
After: Once packet loss starts, it keeps doing continuous expand until the buffer actually fills back up to a healthy level. This prevents the "stuttering" effect during spotty network conditions.
5. Comprehensive testing
Added unit tests for each expand phase and integration tests that verify the decision logic works correctly for expand, accelerate, and preemptive expand scenarios.
Bottom Line
This PR makes audio smoother and less jarring when there's packet loss. Instead of lots of small, noticeable glitches, it does one smooth concealment period until the network recovers.

* reduced overlap size for expand start and expand end
* keep continous expand until buffer level normalizes

### DIFF
--- a/neteq/src/neteq.rs
+++ b/neteq/src/neteq.rs
@@ -527,7 +527,7 @@ impl NetEq {
         }
 
         if self.consecutive_expands > 0 && current_buffer_samples < low_limit as usize {
-            // prefer a single continous expand over many small ones
+            // prefer a single continuous expand over many small ones
             self.consecutive_expands = self.consecutive_expands.saturating_add(1);
             return Ok(Operation::Expand);
         }
@@ -1545,12 +1545,11 @@ mod tests {
             insert_audio(&mut neteq, 10);
         }
 
-        // Consume two frames without adding new audio
-        let _ = neteq.get_audio().unwrap();
-        let _ = neteq.get_audio().unwrap();
-        reset_filtered_level(&mut neteq);
+        // Set target delay to 500ms
+        neteq.delay_manager.set_base_minimum_delay(500);
+        neteq.delay_manager.set_base_maximum_delay(500);
 
-        // Expect PreemtiveExpand,TimeStrectchBuffer,TimeStrectchBuffer a couple times
+        // Expect PreemtiveExpand+TimeStrectchBuffer+TimeStrectchBuffer a couple times
         for _ in 0..3 {
             let _ = neteq.get_audio().unwrap();
             assert_eq!(neteq.last_operation, Operation::PreemptiveExpand);


darioalessandro:
@adam-kiss-sg would you please throw in a couple of tests to make sure that this logic not only does the thing, but continues to do the thing ?

adam-kiss-sg:
@darioalessandro sure, I will take a look at the test for this and the other prs as well, but I can probably only get to this on Friday or next week.

neteq/src/neteq.rs
```diff
@@ -528,49 +546,35 @@ impl NetEq {
...
-        let high_limit = std::cmp::max(target_level_samples, low_limit + 20 * samples_per_ms);
+        // Check for continuous expansion limit (WebRTC uses much higher limits)
+        if self.consecutive_expands > 600 {
+            // ~6 seconds at 10ms frames
+            self.flush();
+            return Ok(Operation::Normal);
+        }
 
-        let buffer_level_samples = self.buffer_level_filter.filtered_current_level();
+        self.consecutive_expands = 0;
```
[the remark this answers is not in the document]
adam-kiss-sg: Not true. This line is right after the check already.

neteq/src/neteq.rs
```diff
@@ -528,49 +546,35 @@ impl NetEq {
...
-            target_level_samples
-                .saturating_sub(K_DECELERATION_TARGET_LEVEL_OFFSET_MS * samples_per_ms),
-        );
-        let high_limit = std::cmp::max(target_level_samples, low_limit + 20 * samples_per_ms);
+        // Check for continuous expansion limit (WebRTC uses much higher limits)
+        if self.consecutive_expands > 600 {
+            // ~6 seconds at 10ms frames
+            self.flush();
+            return Ok(Operation::Normal);
+        }
```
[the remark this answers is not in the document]
adam-kiss-sg: Not true. The reset line is after the check.

adam-kiss-sg:
I moved expand to it's own file and added some tests. Tried to keep it similar to how time_stretch operations work, although it might be a bit overkill to have ExpandFactory and such.

I haven't tested this end-to-end yet, so please don't merge yet :)

darioalessandro:
sgtm!!

adam-kiss-sg:
Did the end to end test, seems to be working fine.

darioalessandro:
Excellent!! I'll test myself tonight!!

neteq/src/neteq.rs
```diff
@@ -505,10 +510,28 @@ impl NetEq {
         // Reset for next frame (matches WebRTC decision_logic.cc:245-246)
         self.timestretch_added_samples = 0;
 
+        // Use WebRTC-style threshold calculations
+        let samples_per_ms = self.config.sample_rate / 1000;
```
darioalessandro, line 514: follow up: since this does not change after config we should cache it.

neteq/src/neteq.rs
```diff
@@ -528,49 +551,35 @@ impl NetEq {
...
 
-        // Calculate thresholds like WebRTC
-        let low_limit = std::cmp::max(
-            target_level_samples * 3 / 4,
-            target_level_samples
-                .saturating_sub(K_DECELERATION_TARGET_LEVEL_OFFSET_MS * samples_per_ms),
-        );
-        let high_limit = std::cmp::max(target_level_samples, low_limit + 20 * samples_per_ms);
+        // Check for continuous expansion limit (WebRTC uses much higher limits)
+        if self.consecutive_expands > 600 {
```
darioalessandro, line 557: @adam-kiss-sg I think this is unreachable due 549 where where check for `if self.consecutive_expands > 0 {`

adam-kiss-sg: You are right. Although, I'm not sure what would be the best fix for this. Line 529:

```
        if self.consecutive_expands > 0 && current_buffer_samples < low_limit as usize {
            // prefer a single continuous expand over many small ones
            self.consecutive_expands = self.consecutive_expands.saturating_add(1);
            return Ok(Operation::Expand);
        }
```

Before this, I noticed a lot of short expands (after a lag, a packet came, resulting in normal (or preemtive expand), then buffer underrun expand again, then normal, then expand. That resulted in some cracking and weird noises. So instead I added the above to wait for enough packets then start playing normally again. Now, if there is a network glitch, there is a short silence then it continues normally.

Now what this will do:
```
        if self.consecutive_expands > 600 {
            // ~6 seconds at 10ms frames
            self.flush();
            return Ok(Operation::Normal);
        }
```

Basically if I add it, what will happen in case of a long lag:
- there will be continuous expands for let's say 20secs, which is basically 20 secs silence (or silent noise)
- then, finally data starts coming, and we have enough to play it normally
- then this check would flush the buffer, attempt to play normally, will find no packets in the buffer (since it was just cleared), and return silence, in the `None` branch in decode_normal.

Maybe something like:
```
    /// Flush the buffer
    pub fn flush(&mut self) {
        self.packet_buffer.flush(&mut self.statistics);
        self.leftover_samples.clear();
        self.reset();
    }

    /// Reset packet statistics and internal state, without clearing incoming buffer
    fn reset(&mut self) {
        self.delay_manager.reset();
        self.buffer_level_filter.reset();
        self.buffer_level_filter
            .set_filtered_buffer_level(self.current_buffer_size_samples());
        self.last_decode_timestamp = None;
        self.consecutive_expands = 0;
        self.packets_received_this_second = 0;
        self.packets_per_sec_snapshot = 0;
        self.last_packets_second_instant = Instant::now();
        self.leftover_time_stretched_samples.clear();
        self.timestretch_added_samples = 0;
    }
```

And then call it like this?
```
        if self.consecutive_expands > 0 {
            // Check for continuous expansion limit (WebRTC uses much higher limits)
            if self.consecutive_expands > 600 {
                // ~6 seconds at 10ms frames
                self.reset();
            }

            self.consecutive_expands = 0;
            return Ok(Operation::ExpandEnd);
        }
```

What do you think?

darioalessandro:
@adam-kiss-sg Thanks for the thoughtful analysis. I went ahead and implemented your proposed approach with a few additional fixes:

**The `reset()` / 600-limit fix (your proposal):**
Applied exactly as you suggested -- the 600-expand check is now inside the `consecutive_expands > 0` block, and `flush()` delegates to a new `reset()` method that preserves the packet buffer. I also added 3 regression tests that verify:
1. The 600-limit actually triggers (it was unreachable before)
2. The system recovers to Normal after the safety valve fires (doesn't re-enter Expand)
3. Sudden disconnects (buffer drops past the ExpandStart window) still recover correctly

**Additional cleanups:**
- Removed `unsafe impl Send for Expand` -- `Expand` only has `usize` + `u64` fields which auto-derive `Send`
- Moved `RNG_STATE` from a global `AtomicU64` to an instance field on `Expand`, so multiple peers don't share/race on the same RNG
- Removed dead `ExpandResult::Error` variant
- Fixed copy-paste comments in tests ("accelerate" -> "expand", "recorder" -> "recorded")

All 68 tests pass. Let me know if you have any concerns with these changes.

PR #477, opened 2025-11-01
allow using neteq wasm without the worker

adam-kiss-sg:
* added a new bin for neteq to allow using WebNetEq without the worker
* changed additional_delay_ms so it can be configured from js

### DIFF
--- a/neteq/src/bin/neteq_wasm.rs
+++ b/neteq/src/bin/neteq_wasm.rs
@@ -1,0 +1,9 @@
+#![cfg_attr(target_arch = "wasm32", no_main)]
+
+#[cfg(target_arch = "wasm32")]
+use neteq as _;
+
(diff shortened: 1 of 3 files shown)


adam-kiss-sg:
Hi!

I wrote my own custom worker for NetEq:
- i have a web worker, which runs WebNetEq
- i have an AudioWorklet, which pulls frames from the web worker (directly, without going through the main thread)
- the main thread is used for setting up the above two, and for sending incoming packets to the worker

If you are interested in this setup, I can share the js code as well.

darioalessandro:
That sounds awesome!! do consider sharing that too!! I'll get this tested and merged asap

adam-kiss-sg:
Pr https://github.com/security-union/videocall-rs/pull/488 contains the audio worklet + worker code I mentioned here.

PR #488, opened 2025-11-14
Neteq worker example

adam-kiss-sg:
Added an example for using NetEQ WASM.

The example includes a server side written in go, that takes an url to
an audio file, downloads the file, converts into 48kHz mono 20ms opus
packets, and returns them over websocket to the client at 20ms intervals.

It also includes a UI to use the above server, with options to:
* configure the source audio file url
* set artificial jitter (on the server side)
* set artificial packet loss (on the server side)
* trigger one-time network drop (this is similar to what often happens 
  when wi-fi or mobile network drops the connection for a short time)
* display for target delay and current buffer size

The example is contained within neteq/examples/worker-example, with the
following components:
* a docker-compose file to start the experience on http://localhost:8080
* main.go file, which contains the server part
* static folder, which contains the JS+html code:
    * static/neteq folder contains the web worker, audio worklet and a 
      small wrapper script for these
    * streaming.js contains code for connecting to the test server
    * index.html + index.js contains html page and it's functions

Additionally, the build within the docker-compose file also builds the
neteq wasm module. The server serves all static files, the wasm files
and the websocket endpoint.

To get started:
```bash
cd neteq/examples/worker-example
docker compose up -d --build && docker compose logs -f
```

### DIFF
--- a/neteq/examples/worker-example/static/index.html
+++ b/neteq/examples/worker-example/static/index.html
@@ -20,14 +20,35 @@
       }
       .grid-container {
         display: grid;
+        align-items: center;
         grid-template-columns: repeat(2, 1fr); /* 2 columns */
-        gap: 10px; /* optional spacing */
+        gap: 12px; /* optional spacing */
+      }
+      .description {
+        font-size: 0.9em;
+        color: #555;
+        display: block;
+      }
+      .input-container {
+        display: flex;
+        align-items: center;
+        margin-bottom: 8px;
+      }
+      .input-description {
+        flex: 1;
+      }
+      .input {
+        flex: 0;
+      }
+      h3.item {
+        grid-column: 1 / -1;
       }
     </style>
   </head>
   <body>
     <h1>Audio Streamer</h1>
     <div class="grid-container" style="max-width: 1000px">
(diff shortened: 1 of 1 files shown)


darioalessandro:
Nice!!

darioalessandro:
This is fantastic, I'll get it tested and merge it!!

darioalessandro:
@adam-kiss-sg this tool is REMARCABLE!!! 

It runs like charm!! I tested the jitter and packet loss features extensively! 

Question, why do you measure Jitter in ms? if jitter is set to 500ms does that mean that all packets have a reception delay of 500ms or what's the probability?

Looks like the pause button is not making audio pause :(,  other than that this is great!

darioalessandro (approved):

adam-kiss-sg:
Hi!

The jitter setting is the maximum delay for any packet. All packets are delayed by a random amount between 0, and the jitter setting, compared to when they would be sent normally. 

I renamed the pause button to "Insert lag spike" to make it more straightforward that it only pauses sending the packets for the time specified.

Also added some descriptions for the inputs in the html page, I tried to make the functionality clearer, please let me know what you think.

Issue #452, opened 2025-10-15
Renaming your user on the client causes you to leave/rejoin the meeting


mgalexander:
**Is your feature request related to a problem? Please describe.**
When a user changes name/screen name we should not drop/rejoin.

**Describe the solution you'd like**
Adjust the name vs resetting or whatever is causing this leave and rejoin pattern.

**Describe alternatives you've considered**
It would be ideal if we don't have to set a different and unique channel, but I have not done research there.

**Additional context**
As observed on every meeting.

iamgabrielsoft:
I'll tackle this also @mgalexander, @testradav @darioalessandro

darioalessandro:
Sgtm

darioalessandro:
Please do a quick design document first, wait for my design doc review please

iamgabrielsoft:
### Bug fix for username rename

When a user changes their display name during an active meeting, the app forces a full page reload, causing the user to loose their connection temporarily. This is because the app does not handle the username change event properly. The change causes yew  navigation to remount the page also causing the disconnection.





### My Solution - 1

The solution is to handle the username change event properly and update the user's display name in the app without causing a full page reload.
1. We need to introduce a stable session id(uuid) that is generated once on join and used for all communication( WebSocket, WebTransport). I think currently, the username/email is used as the session id. So when its changed, videocall-rs views it as a new user and causes a full page reload.

2. Changing the URL triggers Yew router navigation component remount full reconnection. The email parameter serves dual purposes: user identification AND display name

3. Also we can check MediaPacket carries the email, so the new change needs to affect the MediaPacket also.


### My Solution - 2
We can add the user username to the Websocket/WebTransport protocol message type for name update. This will allow local state not to disconnect and also broadcast the name update to all other participants. The client will send **NameUpdate** to the server, the server validates and broadcasts to all participants. Client update their local peer connection list.

@darioalessandro

iamgabrielsoft:
# Diagram

## Current Flow

```
User changes name
       │
       ▼
URL changes: /meeting/old-name/room → /meeting/new-name/room
       │
       ▼
Page reloads → Connection drops → Rejoin as "new user"
       │
       ▼
All participants see the previous name left and the new name joined
```

## Implementation Flow

```
User changes name
       │
       ▼
Send metadata update message (same connection)
       │
       ▼
Server broadcasts new name to all participants
       │
       ▼
All participants see the name update in place
```

## Model Change

```
BEFORE:
[...]

PR #505, opened 2025-12-27
Bug fix user rename

iamgabrielsoft:
This pull request describes and provide solution to bug fix for username renaming for issue #452

### DIFF
--- a/.github/workflows/pr-build-and-deploy.yaml
+++ b/.github/workflows/pr-build-and-deploy.yaml
@@ -139,6 +139,9 @@ jobs:
           - name: web-ui
             dockerfile: Dockerfile.yew
             image: videocall-web-ui
+          - name: dioxus-ui
+            dockerfile: Dockerfile.dioxus
+            image: videocall-dioxus-ui
     steps:
       - name: Checkout PR code
         uses: actions/checkout@v4
@@ -185,6 +188,7 @@ jobs:
             - \`ghcr.io/${{ github.repository_owner }}/videocall-media-server:pr-${prNumber}\`
             - \`ghcr.io/${{ github.repository_owner }}/videocall-meeting-api:pr-${prNumber}\`
             - \`ghcr.io/${{ github.repository_owner }}/videocall-web-ui:pr-${prNumber}\`
+            - \`ghcr.io/${{ github.repository_owner }}/videocall-dioxus-ui:pr-${prNumber}\`
 
             🚀 **Starting deployment...**
 
--- a/.github/workflows/pr-deploy-reusable.yaml
+++ b/.github/workflows/pr-deploy-reusable.yaml
@@ -200,6 +200,7 @@ jobs:
             "videocall-media-server"
             "videocall-meeting-api"
             "videocall-web-ui"
+            "videocall-dioxus-ui"
           )
 
           rm -f /tmp/missing_images_${PR_NUM}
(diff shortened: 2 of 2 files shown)


darioalessandro:
Hey Gabriel just looking at the line count this looks big, please reassess, do you really need all this code to fix the bug? How does your PR work?

iamgabrielsoft:
not really, i tend to test on various cases though. I can revert if you want me to @darioalessandro 

I think the bulk of the diff comes from conflict resolution and code movement, not new logic yet.

dbmate/db/migrations/20250101000000_create_meetings_table.sql
```diff
@@ -0,0 +1,27 @@
+-- Create meetings table to track meeting times
```
darioalessandro: @iamgabrielsoft migrations do not go to this directoy they go to dbmate directory instead

actix-api/src/actors/chat_server.rs
```diff
@@ -330,6 +330,12 @@ async fn send_meeting_info(
     }
 }
 
+// fn handle_subscription_error(e: impl std::fmt::Display, subject: &str) -> String {
```
darioalessandro: commented code, please remove

darioalessandro (commented):
Hey @iamgabrielsoft as always thank you for your hard work.

with that said, the proposed solution goes way beyond of user rename and focuses a lot more into user ownership, approval and other areas of the system.

This MR is extremely complex and bring in changes way outside of user renaming, lets meet soon to discuss

actix-api/src/actors/chat_server.rs
iamgabrielsoft: Acknowledged

Fix

actix-api/src/actors/chat_server.rs
```diff
@@ -481,8 +487,7 @@ mod tests {
 
         assert!(
             result.is_ok(),
-            "JoinRoom with valid user should return Ok, got: {:?}",
-            result
+            "JoinRoom with valid user should return Ok, got: {result:?}"
```
testradav, line 548:
```suggestion
            "Join room with valid user should return Ok. Got: {result:?}"
```

testradav (changes requested):
- [ ] does `Rename-bug-fix.md` need to be part of the code (maybe move to the wiki under an investigations section)
- [ ] meeting owner changes should be part of a separate PR (this PR should be only to fix the user name rename issue)

testradav:
@iamgabrielsoft putting this in draft mode while you work it out.   looking to keep a relatively small open PR list so that it does not get out of hand

darioalessandro:
@iamgabrielsoft if you just change the clients so that when they receive packets with the new username it changes it, wouldnt this simplify everything? 

Now that the backend does not give a crap about usernames it might be the best path forward

iamgabrielsoft:
> @iamgabrielsoft if you just change the clients so that when they receive packets with the new username it changes it, wouldnt this simplify everything?
> 
> Now that the backend does not give a crap about usernames it might be the best path forward

Yeah I notice a gap. I notice late joiners don't see the changed name when their join the call. Even though its changes

iamgabrielsoft:
> Lots of merge conflicts

Oh yes, this is will be factored in :)

iamgabrielsoft:
> > Lots of merge conflicts
> 
> Oh yes, this is will be factored in :)

its ready now @darioalessandro  @testradav

darioalessandro:
@iamgabrielsoft please rebase this PR against against the latest main, this PR is bringing in many unrelated changes, also please consider a solution that is pure client side

iamgabrielsoft:
This was moved ro #720 for reference @darioalessandro

PR #473, opened 2025-10-31
Feature/Host Video in canvas

testradav:
issue #460 

image

PR #517, opened 2026-01-10
feature: self video is a canvas tile

testradav:
self video is part of the canvas tiles with the peers, including hover selectors

- ability to toggle floating bottom right or canvas peers
- remember position of self video (via localStorage/cookie)

- positioned floating bottom right

image

image

- positioned with canvas peers (move to corner button shown):

image

### DIFF
--- a/yew-ui/static/style.css
+++ b/yew-ui/static/style.css
@@ -536,32 +536,26 @@ canvas {
 /* Host camera frame rendered in the bottom right corner */
 .host {
   position: absolute;
-  bottom: 16px; /* Default spacing from bottom */
-  right: 16px; /* Default spacing from right */
-  width: 20%; /* Default width */
-  max-width: 240px; /* Max width to prevent it from being too large on wide screens */
-  min-width: 145px; /* Min width to ensure it's usable, increased to accommodate selectors */
-  /* height: auto; */ /* Let content dictate height with flex */
-  /* aspect-ratio: 16/9; */ /* Aspect ratio might be tricky with added selectors, manage via video element */
-  /* border: 2px solid #444; */
+  bottom: 16px;
+  right: 16px;
+  width: 20%;
+  max-width: 240px;
+  min-width: 145px;
   border-radius: 8px;
-  /* overflow: hidden; */ /* Remove to prevent clipping, flexbox will manage layout */
-  z-index: 15; /* Make sure it's above the grid items if any overlap occurs */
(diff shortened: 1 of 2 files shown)


testradav:
replaces #473

yew-ui/static/style.css
```diff
@@ -543,7 +543,7 @@ canvas {
   min-width: 150px; /* Min width to ensure it's usable, increased to accommodate selectors */
   /* height: auto; */ /* Let content dictate height with flex */
   /* aspect-ratio: 16/9; */ /* Aspect ratio might be tricky with added selectors, manage via video element */
-  border: 2px solid #444;
+  /* border: 2px solid #444; */
```
darioalessandro: nit: commented code, please remove if not needed

darioalessandro (changes requested):
Hey @testradav thanks for the great work here 😄 I do know that some people prefer the "facetime/whatsapp" floating thumbnail vs google meet tile style. 

I found some things I want to bring to your attention:

1. In mobile, the host component occludes the controls:
Image

3. When in floating mode, css should look like it used to, now the tile is wider:

Before:
Image

Now:
Image

4. Controls are cropped when not floating:
Image
I propose that you get rid of the dropdowns there and turn on the settings button always like in mobile:

Image

 
5. Each new icon should go to its own yew component and file. 

Let me know if this makes sense

testradav:
image

- narrow, one attendee (self) floating:
image  

- narrow, one attendee (self) tiled:
image

- narrow/short, two attendees, 1 tiled, self floating:
image

- narrow/short, two attendees tiled:
image

yew-ui/static/style.css
testradav: done https://github.com/security-union/videocall-rs/pull/517/commits/8a6fcf24ddcf22c67d67bab515a2d4f245de01b3

testradav:
> I propose that you get rid of the dropdowns there and turn on the settings button always like in mobile: Image
> 5. Each new icon should go to its own yew component and file.

I will do that as a separate PR.  don't want to compound changes...

darioalessandro:
I had to revert this feature because it caused the video to stop streaming when a new peer joined the call :( https://github.com/security-union/videocall-rs/pull/538 I hope that we can bring this feature back once it does not cause this glitch.

The issue is clear to me, adding and removing host from the dom causes the video encoder to be recreated from scratch which leads to the glitch

PR #519, opened 2026-01-13
Fix GLIBC compatibility and make DATABASE_ENABLED optional

jboyd01:
- Update Dockerfile.actix to use debian:trixie-slim (GLIBC 2.38+)
- Make database pool creation conditional on DATABASE_ENABLED

### DIFF
--- a/actix-api/src/bin/websocket_server.rs
+++ b/actix-api/src/bin/websocket_server.rs
@@ -46,8 +46,8 @@ use sec_api::{
     server_diagnostics::ServerDiagnostics,
     session_manager::SessionManager,
 };
-use tracing::{debug, error, info};
-use videocall_types::truthy;
+use tracing::{debug, error, info, warn};
+use videocall_types::{feature_flags::FeatureFlags, truthy};
 
 const SCOPE: &str = "email profile";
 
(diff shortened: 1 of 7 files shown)


darioalessandro:
@jboyd01 the linter is mad, there's a command in the makefile to trigger format, I think that `cargo fmt` will do the trick.

Also I am establishing a pattern for feature flags, I think db enabled fits right in: https://github.com/security-union/videocall-rs/blob/main/videocall-types/src/feature_flags.rs

actix-api/src/bin/websocket_server.rs
```diff
@@ -342,10 +342,22 @@ async fn main() -> std::io::Result<()> {
         .await
         .expect("Failed to connect to NATS");
 
-    // Create sqlx pool for session management
-    let sqlx_pool = db_pool::create_pool()
-        .await
-        .expect("Failed to create database pool");
+    // Check if database is enabled
+    let db_enabled: bool = truthy(Some(
```
darioalessandro: If you se read_bool_env in the feature flags module you don't have to do this manually.

actix-api/src/webtransport/mod.rs
```diff
@@ -210,10 +210,22 @@ pub async fn start(opt: WebTransportOpt) -> Result<(), Box<dyn std::error::Error
...
+        .to_lowercase() == "true";
+
+    // Create database pool only if enabled
+    let pool = if db_enabled {
+        db_pool::create_pool()
+            .await
+            .expect("Failed to create database pool")
+    } else {
+        // When database is disabled, create a dummy offline pool
+        warn!("DATABASE_ENABLED=false, creating dummy pool that won't be used");
```
darioalessandro: why do you need a dummy pool? just for the typesystem to be happy? you could also make the pool optional?

jboyd01: We need the dummy pool because `ChatServer::new()` and `SessionManager::new()`  both require a `PgPool` parameter in their constructors.   Agreed making it optional would be much cleaner.  Working on it.

jboyd01:
@darioalessandro I believe all issues are addressed.   Not certain if you agree with adding `read_bool_no_prefix()` or if you want to rename the flag DATABASE_ENABLED to FEATURE_DATABASE_ENABLED.  
https://github.com/security-union/videocall-rs/pull/519/files#diff-2ab50ac733fce3b8e7a8c9de52c16981d8b412ce89fbf9a45ab9cc615b2feb9bR112

darioalessandro:
Yeah this works

darioalessandro (approved):
Lgtm

PR #521, opened 2026-01-14
Feature: highlight speaker

ilyatarmanau-lgtm:


videocall-client/src/decode/peer_decode_manager.rs
```diff
@@ -273,6 +280,10 @@ impl Peer {
                     // Check if audio state changed at all
                     let audio_state_changed = self.audio_enabled != metadata.audio_enabled;
 
+                    if !metadata.audio_enabled {
+                                self.is_speaking = false;
+                    }
+
```
testradav, line 286: nit pick:  indent spacing should follow the same as the existing code

```suggestion
                    if !metadata.audio_enabled {
                        self.is_speaking = false;
                    }

```

testradav:
close in lieu of #525

PR #522, opened 2026-01-14
Firefox support

testradav:
firefox v130 or newer supported

image

### DIFF
--- a/Dockerfile.actix
+++ b/Dockerfile.actix
@@ -11,7 +11,7 @@ RUN mkdir -p /app/binaries && \
     -not -name "*.so" -not -name "*.a" -not -name "*.d" -not -name "build_*" \
     -exec cp {} /app/binaries/ \;
 
-FROM debian:bookworm-slim as production
+FROM debian:trixie-slim as production
 
 RUN apt-get --yes update && apt-get --yes install libssl-dev ca-certificates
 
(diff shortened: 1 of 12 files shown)


testradav:
@darioalessandro bump

darioalessandro:
This is a big deal !! I'll review tonight

darioalessandro:
Hey @testradav I am hitting this error while trying to start the mic only in firefox, it has to do with how the mic is initialized probably

Screenshot 2026-01-20 at 23 38 47

testradav:
> Hey @testradav I am hitting this error while trying to start the mic only in firefox, it has to do with how the mic is initialized probably
> 

fixed, but as a fallback solution because firefox doesn't include sampleRate in `MediaTrackSettings.getSettings()` - this is a known browser difference from Chrome. The fix creates a temporary AudioContext when the sample rate isn't available from track settings, using its `sample_rate()` property to detect the system's default audio sample rate (which typically matches the microphone hardware).                

image

@darioalessandro please re-review/merge if acceptable.  otherwise I will put the PR in draft until firefox provides capabilities for sampleRate...

darioalessandro:
@testradav sounds good (No pun intended) reviewing now

darioalessandro:
@testradav I am seeing performance issues on MacOS/Firefox both for audio and video encoders I have to revert this, my bad :( the root-cause is most likely inside MediaStreamTrackProcessor

darioalessandro:
Lol, turns out that the performance issue was on the firefox vp9 encoder, is just too damn slow on macos, so I switched to vp8 ONLY on firefox: https://github.com/security-union/videocall-rs/pull/535

PR #529, opened 2026-01-22
If two phones join with the same name, they can't see each other

anhelinakrutaya-ops:
- Use instance_id for unique user identification

### DIFF
--- a/yew-ui/Cargo.toml
+++ b/yew-ui/Cargo.toml
@@ -40,6 +40,7 @@ reqwasm = "0.5"
 anyhow = "1.0"
 reqwest = { version = "0.12", features = ["json"] }
 urlencoding = "2.1"
+uuid = { version = "1.0", features = ["v4", "js"] }
 
 [dependencies.web-sys]
 version = "0.3.72"
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -38,6 +38,7 @@ use crate::context::{
 use gloo_timers::callback::Timeout;
 use gloo_utils::window;
 use log::{error, warn};
+use uuid;
 use videocall_client::utils::is_ios;
 use videocall_client::{MediaDeviceAccess, VideoCallClient, VideoCallClientOptions};
 use videocall_types::protos::media_packet::media_packet::MediaType;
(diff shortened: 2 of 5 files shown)


darioalessandro:
I’ll take a look this weekend Sent from my iPhoneOn Jan 23, 2026, at 09:07, Antonio Estrada ***@***.***> wrote:﻿

@testradav requested your review on: security-union/videocall-rs#529 If two phones join with the same name, they can&#39;t see each other.

darioalessandro:
## Alternative Approach: Server-Managed Session Identity

Hey! Thanks for digging into this issue. Before we proceed with the current approach, I wanted to share some thoughts on a more scalable solution.

### The Core Problem

The current approach concatenates `username-uuid` on the client side, then parses it apart for display. This works, but it's treating the symptom rather than the root cause.

The root cause is that we're using the display name as the connection identity. Looking at how systems like Google Meet handle this:

| Concept | Google Meet | videocall-rs (current) |
|---------|-------------|------------------------|
| Identity (routing/tracking) | Account ID or session ID | Username |
| Display name | Profile name (cosmetic) | Same as identity |
| Uniqueness | Server-enforced | Client-provided |

When identity and display name are the same field, collisions are inevitable.

### What We Already Have

Here's the interesting part—the server already generates a unique session ID for every connection:

```rust
// WebSocket (chat_session.rs)
let session_id = Uuid::new_v4().to_string();

// WebTransport (webtransport/mod.rs)
let session_id = uuid::Uuid::new_v4().to_string();
```

It's just not being used for routing. The infrastructure for proper identity management already exists.

### Proposed Direction

Instead of client-side UUID generation, we could:

1. **Use the existing server-generated `session_id` for routing**
   - NATS subject: `room.{room_id}.{session_id}` instead of `room.{room_id}.{username}`
   - This guarantees uniqueness without client involvement

2. **Treat display name as metadata, not identity**
   - Client sends display name after connecting
   - Server broadcasts it with `PARTICIPANT_JOINED` events
   - Multiple users with the same display name is fine—they're tracked by session_id

3. **Simplify the URL**
   - Could be just `/lobby/{room_id}` since identity is server-assigned

This approach:
- Removes the need for the `uuid` dependency on the frontend
- Eliminates the `extract_username_from_userid()` parsing throughout the codebase
- Follows the same pattern production systems use
- Scales better as the codebase grows

### Tradeoffs

This is more work than the current PR—it touches the URL structure, NATS routing, database schema, and protocol. But it solves the problem at the right layer and won't need revisiting.

What do you think?

anhelinakrutaya-ops:
Hey Dario! Thanks for your thoughts. I agree with your approach.

Initially, I also considered using `session_id` for uniqueness and even tried implementing it, but I ran into an issue. Since we create `session_id` independently in both `webtransport_server` and `websocket_server`, and both publish to NATS, during the election phase (before the best protocol is selected) we end up with two sessions for the same user being visible at the same time. Once the protocol is chosen, one of them disappears. I tried filtering on the client to display only a single session, but that didn’t work reliably. It’s possible I missed something there.
 
But I’ve also been thinking about an alternative solution, though it would require more backend changes:

- Generate `session_id` once per user (not per connection)
- Pass `session_id` in the `MEETING_STARTED` packet
- Use the same `session`_id for both WebSocket and WebTransport connections
- Use `session_id` for NATS routing (`room.{room_id}.{session_id}`) instead of username

With this approach, both connections would share the same identity in NATS routing, which should eliminate the duplicate session issue during the election phase.

What do you think about this direction?

darioalessandro:
@anhelinakrutaya-ops that makes sense, how would you generate the session_id once?

anhelinakrutaya-ops:
Good question.
I was considering modifying `session_id`, but I came to the conclusion that it’s better not to touch it at all.

If common authentication is planned for the future (as I see currently Google auth exists only on the WebSocket server), it would make more sense to introduce a unique user_id as the primary identity instead of relying on usernames. That would likely require a separate authentication service anyway.

As a temporary solution, I suggest that before any WebSocket or WebTransport connections are created, the client makes a single HTTP request to obtain a unique `user_id / instance_id`. This request would happen right after the user clicks Join / Create, and before ConnectionManager starts establishing connections. The server generates and returns a UUID-based `user_id / instance_id`, which is then passed to both WebSocket and WebTransport connection URLs, for example:
`ws://server/lobby/{email}/{room_id}/{instance_id}`

darioalessandro:
@anhelinakrutaya-ops I put some thought into this, how about:


# | Requirement
1 | Server generates session_id on connection
2 | Server sends session_id to client (new message or first RTT response)
3 | Client includes session_id in all outgoing MediaPackets
4 | Server tracks Testing → Active state, broadcasts to NATS only on Active
5 | Client delays HEARTBEAT until after election
6 | Receiving clients use session_id as peer key, email as display name
7 | Add session_id field to MediaPacket protobuf

anhelinakrutaya-ops:
Regarding point 1, “Server generates session_id on connection”: currently it seems there are two servers - `websocket_server` and `webtransport_server`. Which one is supposed to generate the `session_id`?
Will each server generate `session_id` independently depending on which one the client happens to connect to first, or should there be a separate service responsible for session/connection management?

Or do you mean that we keep generating the session_id on each server as it is and add a delay until after the election?

darioalessandro:
Yeah that is kind of a flaw, it is ok if we have some duplication for now. Each server should generate its own uiid hopefully we can consolidate that soon

testradav:
relates to #567

Issue #465, opened 2025-10-28
After disabling screenshare, the peer screen did not turn full fullscreen.


darioalessandro:
**Describe the bug**
Since there's only 1 peer connected, the image should pin to the screen.

**To Reproduce**
1. get on a call 1:1, then have one of the peers share screen. 
2. have that peer stop sharing.

**Expected behavior**
The remaining peer should be full screen but it is not. 

**Desktop (please complete the following information):**
 Brave

PR #530, opened 2026-01-23
fix problem when user cancel share screen and After disabling screenshare, the peer screen did not turn full fullscreen. #465

Alena-Melnikava:
Prepare fix for several small bugs:
-  Bug with share screen button: when user press share screen button and then cancel window where should select what to share, screen share button remains on;
- After disabling screenshare, the peer screen did not turn full fullscreen;
-  bug when browser bar with stop sharing button is not always disactivated when user stop share of screen.

### DIFF
--- a/videocall-client/src/encode/screen_encoder.rs
+++ b/videocall-client/src/encode/screen_encoder.rs
@@ -157,7 +157,7 @@ impl ScreenEncoder {
     /// The user is prompted by the browser to select which window or screen to encode.
     ///
     /// This will toggle the enabled state of the encoder.
-    pub fn start(&mut self) {
+    pub fn start(&mut self, screen_stream: MediaStream) {
         let EncoderState {
             enabled,
             destroy,
@@ -169,6 +169,7 @@ impl ScreenEncoder {
         enabled.store(true, Ordering::Release);
         destroy.store(false, Ordering::Release);
 
+        let screen_to_share = screen_stream;
         let client = self.client.clone();
         let userid = client.userid().clone();
         let aes = client.aes();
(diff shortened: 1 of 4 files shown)


darioalessandro:
Wow thank you for your contribution! Will take a look this weekend Sent from my iPhoneOn Jan 23, 2026, at 10:32, Alena-Melnikava ***@***.***> wrote:﻿Prepare fix for several small bugs:

Bug with share screen button: when user press share screen button and then cancel window where should select what to share, screen share button remains on;
After disabling screenshare, the peer screen did not turn full fullscreen;
bug when browser bar with stop sharing button is not always disactivated when user stop share of screen.




You can view, comment on, or merge this pull request online at:
  https://github.com/security-union/videocall-rs/pull/530

Commit Summary

  1311f98  fix problem when user cancel share screen


File Changes  (4 files)

  
    M
    videocall-client/src/encode/screen_encoder.rs
    (23)
  
  
    M
    yew-ui/src/components/attendants.rs
    (119)
  
  
    M
    yew-ui/src/components/host.rs
    (13)
  
  
    M
    yew-ui/src/components/video_control_buttons.rs
    (15)
  


Patch Links:

  https://github.com/security-union/videocall-rs/pull/530.patch
  https://github.com/security-union/videocall-rs/pull/530.diff

Alena-Melnikava:
I agreed with almost all comments excluding description of logic in architecture concern.
I understood your logic of architecture, but we have a small problem here:
According to my investigation, the biggest part of screen view updates are activated before we call videocall-client. Because of this, we should update screen view before user finally activates option (how it happened with screen share button) or screen view update is not activated in time (how it was with situaton when user switch off screen share, but other users screens aren't return to full view). I can try to save call of get_display_media() in videocall-client, but I afraid that situations with small problems connected with updates of screen not in time will continue occur.

darioalessandro:
Thanks for explaining the timing concern - that makes sense! You're right that the UI needs to react at several points during the screen share flow.

The good news is we can solve this while keeping `get_display_media()` in `videocall-client`. The solution is to have the encoder notify the UI at key moments via callbacks. Here's how it could work:

```rust
// In videocall-client's ScreenEncoder

pub struct ScreenShareCallbacks {
    pub on_requesting: Box<dyn Fn()>,      // Called when dialog opens
    pub on_started: Box<dyn Fn()>,          // Called when stream acquired  
    pub on_cancelled: Box<dyn Fn()>,        // Called when user cancels dialog
    pub on_stopped: Box<dyn Fn()>,          // Called when browser stops sharing
}

pub async fn start(&mut self, callbacks: ScreenShareCallbacks) {
    (callbacks.on_requesting)();  // UI can now show "requesting" state
    
    match self.get_display_media().await {
        Ok(stream) => {
            // Set up onended handler that calls callbacks.on_stopped
            (callbacks.on_started)();
        }
        Err(_) => {
            (callbacks.on_cancelled)();  // UI resets the button
        }
    }
}
```

Then in `yew-ui`, the code becomes simpler - just pass callbacks:

```rust
self.screen.start(ScreenShareCallbacks {
    on_requesting: || { /* update button to requesting state */ },
    on_started: || { /* update button to active */ },
    on_cancelled: || { /* reset button */ },
    on_stopped: || { /* reset button and layout */ },
});
```

This gives you:
- The same timing control you need for UI updates
- All media logic stays encapsulated in `videocall-client`
- The UI remains thin and focused on presentation

Would this approach work for your use case?

darioalessandro:
HI @Alena-Melnikava I took a stab at this PR, let me know what you think https://github.com/security-union/videocall-rs/pull/546

Alena-Melnikava:
Corrections should be reworked with accordance of changes which were done by Dario for issue #546

PR #532, opened 2026-01-27
feature: enhance MediaStreamTrackProcessor polyfill performance using workers 

darioalessandro:
### Background

Firefox doesn't support native `MediaStreamTrackProcessor` (as of Firefox 143). The polyfill works by:
1. Capturing video frames via canvas
2. Creating `VideoFrame` objects
3. Feeding them to the WebCodecs encoder

The bottleneck is the GPU → CPU → GPU roundtrip when copying frames through canvas. While we can't eliminate this (it's a browser limitation), we can reduce main thread blocking.

### What This PR Does

**Before:** All frame processing happened on the main thread, causing UI jank during video calls.

**After:** 
- `VideoFrame` creation is offloaded to a Web Worker
- Uses `createImageBitmap()` instead of `canvas.drawImage()` (potentially more optimized)
- Falls back to canvas approach if Workers unavailable

### Code Paths

| Browser | API Available | Path Used |
|---------|---------------|-----------|
| Chrome/Edge/Safari | Native `MediaStreamTrackProcessor` | Native (no polyfill) |
| Firefox 130+ | `requestVideoFrameCallback` + Workers | Worker-based polyfill |
| Older Firefox | `requestAnimationFrame` + Workers | Worker-based polyfill (RAF fallback) |
| No Worker support | Canvas only | Canvas-based polyfill |

### Performance Impact

- **Main thread:** Less blocking, smoother UI
- **Frame latency:** Similar (GPU readback is still the bottleneck)
- **Overall:** Better perceived performance in Firefox

### Known Limitations

- Video encoding in Firefox will always be slower than Chrome/Safari until Firefox ships native `MediaStreamTrackProcessor`
- Mozilla announced support for "early 2025" but it hasn't landed yet

### Future Work

- [ ] Refactor polyfill into separate JS files for better maintainability
- [ ] Add performance metrics/logging
- [ ] Consider lower resolution/framerate defaults for Firefox

### DIFF
--- a/videocall-client/src/encode/microphone_encoder.rs
+++ b/videocall-client/src/encode/microphone_encoder.rs
@@ -278,35 +278,39 @@ impl MicrophoneEncoder {
 
             // Sample Rate hasn't been added to the web_sys crate
             // Firefox doesn't report sampleRate in MediaTrackSettings, so we need a fallback
-            let input_rate: u32 =
-                match js_sys::Reflect::get(&track_settings, &JsValue::from_str("sampleRate")) {
-                    Ok(v) => match v.as_f64() {
-                        Some(f) => f as u32,
-                        None => {
-                            // Firefox fallback: create a temporary AudioContext to get system sample rate
-                            log::info!("sampleRate not in track settings (Firefox), using AudioContext default");
-                            match AudioContext::new() {
-                                Ok(temp_ctx) => {
-                                    let rate = temp_ctx.sample_rate() as u32;
-                                    let _ = temp_ctx.close();
-                                    rate
+            let input_rate: u32 = match js_sys::Reflect::get(
+                &track_settings,
+                &JsValue::from_str("sampleRate"),
+            ) {
+                Ok(v) => match v.as_f64() {
+                    Some(f) => f as u32,
+                    None => {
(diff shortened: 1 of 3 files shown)


videocall-client/src/encode/microphone_encoder.rs
```diff
@@ -278,35 +278,39 @@ impl MicrophoneEncoder {
 
             // Sample Rate hasn't been added to the web_sys crate
             // Firefox doesn't report sampleRate in MediaTrackSettings, so we need a fallback
-            let input_rate: u32 =
```
darioalessandro, line 281: this is just `cargo fmt`

yew-ui/src/context.rs
```diff
@@ -82,7 +82,10 @@ pub fn load_self_video_position_from_storage() -> bool {
 /// Persist the self-video position preference to `localStorage`.
 pub fn save_self_video_position_to_storage(is_floating: bool) {
     if let Some(storage) = web_sys::window().and_then(|w| w.local_storage().ok().flatten()) {
-        let _ = storage.set_item(SELF_VIDEO_POSITION_KEY, if is_floating { "true" } else { "false" });
+        let _ = storage.set_item(
```
darioalessandro, line 85: `cargo fmt`

Issue #299, opened 2025-06-29
App blows up if you do not have a connected camera AND or microphone


darioalessandro:
The only feature question that I did not process correctly reading before playing was that my chrome client app blows up if I don't have a camera device 

Error requesting permissions: Please make sure to allow access to both camera and microphone. (JsValue(NotFoundError: Requested device not found
undefined))

plugged in a camera and presto it works great.

testradav:
@darioalessandro is this still an open issue? (eg. app breaks if no camera detected)

PR #539, opened 2026-01-27
App blows up if you do not have a connected camera AND or microphone #299

Alena-Melnikava:
Added check if user activated mic or video usage. If mic and video usage weren't activated, we can connect to meeting session without check of media device status. 
Media device status is checked only when user try to activate mic or video or both of them usage.

### DIFF
--- a/.github/workflows/cargo-test.yaml
+++ b/.github/workflows/cargo-test.yaml
@@ -1,6 +1,6 @@
 # Based on https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md
 
-on: 
+on:
   pull_request:
     paths:
       - 'actix-api/**'
@@ -11,33 +11,19 @@ on:
       - 'neteq/**'
       - 'videocall-client/**'
 name: check backend
-  
+
 jobs:
   test:
     name: cargo tests_run
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
-      # Cache Docker layers
-      - name: Cache Docker layers
-        uses: actions/cache@v3
-        with:
-          path: /tmp/.docker
-          key: ${{ runner.os }}-docker-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-docker-
-
-      # Cache Rust build artifacts
-      - name: Cache Cargo Registry
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-
+      - name: Clean up Docker resources
+        run: docker system prune -f --volumes
 
       - run: make tests_run
+
+      - name: Teardown
+        if: always()
+        run: make tests_down
--- a/.github/workflows/wasm-test.yaml
+++ b/.github/workflows/wasm-test.yaml
@@ -8,6 +8,7 @@ on:
       - 'videocall-codecs/**'
       - 'videocall-types/**'
       - 'protobuf/**'
+      - 'yew-ui/**'
   pull_request:
     paths:
       - 'videocall-client/**'
@@ -15,6 +16,7 @@ on:
       - 'neteq/**'
       - 'videocall-types/**'
       - 'protobuf/**'
+      - 'yew-ui/**'
 
 jobs:
   wasm-test:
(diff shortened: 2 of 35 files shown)


darioalessandro:
That is Wonderful, looking now!!

darioalessandro:
Hey @Alena-Melnikava !! thanks for working on this!! 

This is how it looks now when you join a call:

Screenshot 2026-01-29 at 22 33 31

all drop downs are empty which is a bit weird.

Lets just follow the pattern that Google Meets uses:

When you join a meeting and have never allowed access to the devices this is what you get

Screenshot 2026-01-29 at 22 32 02

If you refuse access, this is how the meeting looks:

Screenshot 2026-01-29 at 22 33 02

If you click on the camera then you get this dialog:

Screenshot 2026-01-29 at 22 33 06

If you do not give any permissions, then you can still listen in using the default system audio output but you can't change it.

For the dialogs just use the existing components look and feel for alerts, we will add nice graphics after

darioalessandro:
@Alena-Melnikava any updates here?

Alena-Melnikava:
I just finished some updates for my issue about small problems of screenshare usage and return back to these issue. I have started investigation how to introduce UI information of user about problems with his media devices.

Alena-Melnikava:
@darioalessandro 
I looked again through changes in UI which you proposed and try to look at them as if I user of videocall-rs. And I want to add some comments and suggestions from my side. 

As I understood the first information message in Google meet you get during attemp to join to meeting room. But, unlike Google meet, videocall-rs doesn't provide user with option switch on mic or/and camera before he join to meeting room. I think in the following situation such message screen can confuse user: I do nothing and even can't do anything, but already get message which provoke me switch on mic or/and camera. I can be not ready for such!

As alternative I can propose the following behavior of UI:
1. When user will join to meeting room, we will check the status of his mic and camera. If both devices accessible then no any special messages. In case of something or both devices are not accessible we will show for user message screen with information that he has problems with mic or/and camera - we pay attention of user that something is wrong because maybe user doesn't know about problem (maybe his camera or mic is attached but doesn't work).
2. On videocall-rs screen the problem device (mic, camera or both) will be marked with "!" symbol on the button to show user with which device he has problem. We also can like disable button of problem device: button will be on the screen but user can't push to it - to don't provoke problem situations when user try to activate device which is not accessible.
3. If user attach another device and switch to it, we again check if device is accessible, and appropriate button for device become accessible without any additional marks if everything is OK.

What do you think about such behavior?

darioalessandro:
Hey @Alena-Melnikava here are my comments:

```As I understood the first information message in Google meet you get during attemp to join to meeting room. But, unlike Google meet, videocall-rs doesn't provide user with option switch on mic or/and camera before he join to meeting room. I think in the following situation such message screen can confuse user: I do nothing and even can't do anything, but already get message which provoke me switch on mic or/and camera. I can be not ready for such!``` 

Dario: the idea is that prior to joining the call we show the banner, "Do you want people to hear you in the meeting" it is a very clear message imo, we don't have to super original with the wording, Meet does a pretty good job explaining the problem and the possible solutions.


```As alternative I can propose the following behavior of UI:```

```
When user will join to meeting room, we will check the status of his mic and camera. 

If both devices accessible then no any special messages. In case of something or both devices are not accessible we will show for user message screen with information that he has problems with mic or/and camera - we pay attention of user that something is wrong because maybe user doesn't know about problem (maybe his camera or mic is attached but doesn't work).```

Dario: On top of the initial banner prior to joining the call, yes we should absolutely check the access to the devices after they join. It is important to mention that the first time, the app will never have the permissions granted to access mic and camera so that banner will almost always be shown if they "ignore" the first banner prior to joining the call.

When you say problem device I do not really follow, to me it sounds like a device that we have access to but we can't use for some reason, to me that is another edge case, we do handle that for the mic, we show an error if you try to start using a mic and then the mic_encoder.rs throws an error.

```
On videocall-rs screen the problem device (mic, camera or both) will be marked with "!" symbol on the button to show user with which device he has problem. We also can like disable button of problem device: button will be on the screen but user can't push to it -`
```
Dario: I do think that it would be a non-trivial amount of work to remember devices that you have access to but there's a problem using, I would leave that out of scope and only show the ! symbol as a button when you do not have access to either mic or cameras (or both) and clicking on them should result in showing the banner prompting them to allow access to the specific device.

``` to don't provoke problem situations when user try to activate device which is not accessible.
```

```
If user attach another device and switch to it, we again check if device is accessible, and appropriate button for device become accessible without any additional marks if everything is OK.
```

Dario: Yeah we currently do monitor for new devices, assuming that the user gives you permission to MediaDevices we do get the notification here:
https://github.com/security-union/videocall-rs/blob/0b74421315d4fa705020530582f50b8b4f22b299/videocall-client/src/media_devices/media_device_list.rs#L441

Dario: we also should show an error if the user just does not have any devices available for audio, video or speaker.

I hope it makes sense.

PR #545, opened 2026-01-29
Bug fix: video and mic turn off on peer joining

testradav:
Summary of the fix

The root cause was in `CameraEncoder::start()`. When the Host component gets recreated (which happens when a new peer joins), the following sequence occurred:

1. camera.select(device_id) is called → sets switching=true
2. After 1 second, camera.start() is called
3. The encoding loop immediately exits because switching=true
 
The fix:
Reset the switching flag at the beginning of start() before entering the encoding loop. This ensures a fresh start doesn't get affected by a previous select() call

@darioalessandro I created this branch on top of my `feature/self-video` branch, but I don't think git will reapply those changes that were reverted.  How is that reapplied? (eg. revert the revert)

### DIFF
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -871,6 +871,7 @@ impl Component for AttendantsComponent {
                                     is_connected={self.client.is_connected()}
                                     is_floating={self.self_video_floating}
                                     on_position_toggle={ctx.link().callback(|_| UserScreenToggleAction::SelfVideoPosition)}
+                                    on_screen_share_state={ctx.link().callback(Msg::ScreenShareStateChange)}
                                 />
                             }
                         } else {


videocall-client/src/encode/camera_encoder.rs
```diff
@@ -327,6 +333,33 @@ impl CameraEncoder {
             let mut current_encoder_height = height as u32;
 
             loop {
+                // Check the video element state and re-apply srcObject if needed.
+                // This handles cases where the video element is recreated during re-render.
+                if let Some(current_video_elem) = window()
```
darioalessandro, line 338: Querying the dom like this will be expensive, this callback gets called around 100hz! I'll try to optimize something in the https://github.com/security-union/videocall-rs/tree/video-on-canvas-integration-branch branch

yew-ui/src/components/host.rs
```diff
@@ -120,6 +121,18 @@ pub struct MeetingProps {
     #[prop_or_default]
     pub on_microphone_error: Callback<String>,
 
+    /// Whether the client is connected to the server
+    #[prop_or_default]
+    pub is_connected: bool,
```
darioalessandro, line 126: videocall client already has this property

PR #546, opened 2026-01-30
Add screen share state callback to videocall-client

darioalessandro:
Adds a callback mechanism in `ScreenEncoder` to notify the UI about screen share state changes.

## Problem

Previously, the UI had no way to know when:
- User cancelled the browser screen picker
- User stopped sharing via browser's "Stop sharing" button

## Solution

Added `ScreenShareEvent` enum with three states:
- `Started` - screen share successfully started
- `Cancelled` - user dismissed the browser picker
- `Stopped` - screen share ended

The `ScreenEncoder` now accepts an `on_state_change` callback and emits these events appropriately, allowing the UI to react without managing MediaStream lifecycle directly.

### DIFF
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -83,7 +83,7 @@ pub enum Msg {
     OnPeerRemoved(String),
     OnFirstFrame((String, MediaType)),
     OnMicrophoneError(String),
-    DismissMicError,
+    DismissUserError,
     UserScreenAction(UserScreenToggleAction),
     #[cfg(feature = "fake-peers")]
     AddFakePeer,
@@ -148,7 +148,8 @@ pub struct AttendantsComponent {
     pub device_settings_open: bool,
     pub error: Option<String>,
     pub encoder_settings: Option<String>,
-    pub mic_error: Option<String>,
+    /// Generic user-visible error message shown in a dialog
+    pub user_error: Option<String>,
     pending_mic_enable: bool,
     pending_video_enable: bool,
     pending_screen_share: bool,
@@ -386,7 +387,7 @@ impl Component for AttendantsComponent {
             device_settings_open: false,
             error: None,
             encoder_settings: None,
-            mic_error: None,
+            user_error: None,
             pending_mic_enable: false,
             pending_video_enable: false,
             pending_screen_share: false,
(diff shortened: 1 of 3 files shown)


yew-ui/src/components/attendants.rs
```diff
@@ -663,6 +666,20 @@ impl Component for AttendantsComponent {
                 self.meeting_ended_message = Some(end_time);
                 true
             }
+            Msg::ScreenShareStateChange(event) => {
```
darioalessandro, line 670: @Alena-Melnikava this is what I meant ^^

Alena-Melnikava: I'm still working under the changed logic of my correction: get_display_media() will be called in videocall-lient and yew-ui will get only the state of screen share. The main logic already works, but I still have some small things which should be fixed:
- Sharing for users which are connected to meeting session is started earlier than owner of sharing select what he want to share.
- rework logic which process stop sharing button in status record in browser to process it correctly

darioalessandro: Hey @Alena-Melnikava I wont do this again, but I got pretty far on this PR and would like to use this instead. 

I apologize because it is frustrating when someone takes the PR that I am working on.

yew-ui/src/components/host.rs
```diff
@@ -253,7 +253,8 @@ impl Component for Host {
         // TODO: use atomic bools for the encoders
         self.client.set_audio_enabled(self.mic_enabled);
         self.client.set_video_enabled(self.video_enabled);
-        self.client.set_screen_enabled(self.share_screen);
+        // Note: set_screen_enabled is now handled internally by ScreenEncoder
```
darioalessandro, line 256: @Alena-Melnikava self.client is a very leaky abstraction right now, I am trying to reduce the calls from the ui.

darioalessandro: at some point I am hoping to remove all this calls from here and into videocall-client

yew-ui/src/components/attendants.rs
Alena-Melnikava: Oh, sorry. You marked this idea in comments for my changes and I thought that I should rework my changes in accordance with it.  I didn't plan to take issue with which you work.

PR #549, opened 2026-02-03
fix: move encoders to context to survive Host recreation

darioalessandro:
## Summary

- Move `CameraEncoder`, `MicrophoneEncoder`, and `ScreenEncoder` ownership from `Host` component to a new `MediaEncoderCtx` context
- Encoders now survive `Host` component recreation (e.g., when switching between floating/embedded self-video)
- Remove DOM polling hack from `camera_encoder.rs` that was checking/re-applying video element state every frame

## Problem

The `Host` component was being recreated when switching between floating and embedded self-video positions. Since `Host` owned the encoders, they were destroyed and recreated, causing video to stop.

PR #545 added a DOM polling workaround that checked the video element state on every frame and re-applied the stream if needed. This was a band-aid fix.

## Solution

1. Created `MediaEncoderCtx` in `context.rs` that holds encoders in `Rc<RefCell<MediaEncoderInner>>`
2. `AttendantsComponent` creates and provides `MediaEncoderCtx` via `ContextProvider`
3. `Host` gets encoders from context and initializes them only once (checks `is_initialized()`)
4. `Host::destroy()` no longer stops encoders - they persist across Host recreation
5. `AttendantsComponent::destroy()` calls `stop_all()` when meeting ends
6. Removed the 27-line DOM polling hack from `camera_encoder.rs`

## Test Plan

- [ ] Video/audio work normally when joining a call
- [ ] Switching between floating and embedded self-video doesn't stop video
- [ ] Peer joining doesn't stop your video
- [ ] Device switching (camera/mic) still works
- [ ] Leaving meeting properly stops encoders (check console for "encoders stopped" log)

### DIFF
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -1087,4 +1087,10 @@ impl Component for AttendantsComponent {
             </ContextProvider<MeetingTimeCtx>>
         }
     }
+
+    fn destroy(&mut self, _ctx: &Context<Self>) {
+        // Stop all encoders when meeting ends
+        self.media_encoder_ctx.stop_all();
+        log::info!("AttendantsComponent destroyed, encoders stopped");
+    }
 }
--- a/yew-ui/src/context.rs
+++ b/yew-ui/src/context.rs
@@ -8,7 +8,6 @@
 use std::cell::RefCell;
 use std::rc::Rc;
 use videocall_client::{CameraEncoder, MicrophoneEncoderTrait, ScreenEncoder, VideoCallClient};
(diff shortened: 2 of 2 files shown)


yew-ui/src/components/attendants.rs
```diff
@@ -1000,24 +1006,6 @@ impl Component for AttendantsComponent {
                                             }
                                         } else { html!{} }
                                     }
-                                    {
-                                         if media_access_granted {
-                                             html! {<Host
```
darioalessandro, line 1005: this was just unused

PR #552, opened 2026-02-04
Fix camera switching when device changes

palina-piliushyna:
Fixes an issue where switching camera devices did not correctly update
the encoder state.

Changes:
- handle camera device change more reliably
- update encoder state on device switch


darioalessandro:
@palina-piliushyna @testradav  I am going to merge this PR with https://github.com/security-union/videocall-rs/pull/552 because they tackle similar issues and call it: 

bug: fix device selection

PR #550, opened 2026-02-03
feature/meeting-ownership

testradav:
- Implemented meeting ownership 
- Implemented waiting room capability

## Documentation and feature flow

https://github.com/testradav/videocall-rs/blob/ef8c3e9a919c0d5192d7ae056aa9259571fd52e7/docs/MEETING_API.md

https://github.com/testradav/videocall-rs/blob/ef8c3e9a919c0d5192d7ae056aa9259571fd52e7/docs/MEETING_OWNERSHIP.md

### Shows the meetings you have created with ability to delete them. Other users will not be able to use the same meeting ID that you have reserved unless you delete the meeting first.
image

### create or join a meeting button shows up only when there's data in the meeting ID field
image

### active meetings show attendee and time stats
image

### waiting room message to host or any anyone else that has already been allowed in the meeting
image

### multiple people waiting to enter a meeting or display in a consolidated list with ability to accept all or one by one
image

### wait dialogue to be allowed into the meeting
image

### host identified by the word host in parentheses
image

image

### attempting to join a meeting that is not owned by you and the owner has not started it will display unable to join message
image

PR #559, opened 2026-02-05
Fix/ci cargo registry cache

iamgabrielsoft:
- Remove unused host-level Cargo registry and Docker layer caches that were causing
  `allocator-api2 .cargo-ok` corruption errors in CI
- Add Docker cleanup step before tests to ensure clean state
- Add teardown step to clean up containers/volumes after tests
- Update `actions/checkout` from v3 to v4

## Context
The CI was caching `~/.cargo/registry` and `~/.cargo/git` on the GitHub Actions
runner host, but `make tests_run` executes inside Docker containers with their own
named volumes. The host cache was never used by the containers and became corrupted
over time, causing failures like:

### DIFF
--- a/actix-api/src/actors/chat_server.rs
+++ b/actix-api/src/actors/chat_server.rs
@@ -481,8 +481,7 @@ mod tests {
 
         assert!(
             result.is_ok(),
-            "JoinRoom with valid user should return Ok, got: {:?}",
-            result
+            "JoinRoom with valid user should return Ok, got: {result:?}"
         );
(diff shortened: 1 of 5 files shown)


iamgabrielsoft:
This keeps happening due to cargo caches in docker. Re-running the job fixes it, but the corruption is intermittent

darioalessandro:
Ok chief thanks I'll take e look soon

testradav:
relates to #550 tests

PR #560, opened 2026-02-05
Fix problems which connected with sharescreen functionality

Alena-Melnikava:
Fix for the following problems:
- Screen share buttion is activated too early, before peer select what he want to share;
- MediaStream doen't stop when screenshare button pressed second time (disactivated);
- After screen share is stopped, peer screen did not turn full fullscreen.

### DIFF
--- a/.github/workflows/cargo-test.yaml
+++ b/.github/workflows/cargo-test.yaml
@@ -1,6 +1,6 @@
 # Based on https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md
 
-on: 
+on:
   pull_request:
     paths:
       - 'actix-api/**'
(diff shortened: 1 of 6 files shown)


darioalessandro:
Hi @Alena-Melnikava are you able to reproduce the is on the latest master?

Alena-Melnikava:
Yes, Antonio merged your changes to our main branch and I tested behavior of screen share with them. On the result of the tests I prepared my changes.

darioalessandro:
@Alena-Melnikava I merged this change into an integration branch and I am doing some tweaks there

yew-ui/src/components/attendants.rs
```diff
@@ -45,6 +45,13 @@ use web_sys::*;
 use yew::prelude::*;
 use yew::{html, Component, Context, Html};
 
+#[derive(Clone, PartialEq)]
+pub enum ScreenShareState {
```
darioalessandro, line 49: I think this is a good pattern, it replaces the `share_screen` bool, I'll finish the cleanup in the integration branch

yew-ui/src/components/attendants.rs
```diff
@@ -521,13 +530,10 @@ impl Component for AttendantsComponent {
                 match action {
                     MeetingAction::ToggleScreenShare => {
                         if !self.share_screen {
-                            if self.media_device_access.is_granted() {
```
darioalessandro, line 524: @Alena-Melnikava I am bringing this back in the integration branch

darioalessandro: actually this is not needed according to https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia

yew-ui/src/components/attendants.rs
Alena-Melnikava: Thanks.

yew-ui/src/components/attendants.rs
Alena-Melnikava: According my investigation, we don't need check of media devices.

PR #566, opened 2026-02-08
Fix device selection

darioalessandro:
## Summary

Fixes unreliable device selection for camera, microphone, and speakers. Merges and fixes issues from #552 and #557.

### 1. Camera and microphone always fell back to the default device

Both `CameraEncoder` and `MicrophoneEncoder` used `set_device_id(&device_id.into())` when calling `getUserMedia`. The browser treats a bare `deviceId` as an **ideal** (preferred) constraint — if the requested device isn't available, it silently falls back to the default. Changed both to use `{ exact: device_id }`, which forces the browser to use the specific device or fail explicitly.

### 2. Race condition: two encoding loops running concurrently on camera switch

The `VideoDeviceChanged` handler called `stop()` (which clears `enabled`), then `select()` (which checks `enabled` to decide whether to set `switching`). Since `enabled` was already false, `switching` was never set, and the old encoding loop continued running alongside the new one.

Fixed by using the same `switching`-based pattern as `AudioDeviceChanged`: call `select()` while the camera is still enabled so the `switching` flag gets set, causing the old loop to exit. A 1-second timeout then restarts the encoder.

### 3. Device change handler read wrong device, broke selection on hot-plug

When `ondevicechange` fired (USB mic plugged in, etc.), the handler read `devices.first()` as the "current selection" instead of the actual selected device. This caused incorrect fallback behavior — the user's real selection could be silently lost.

Fixed by sharing the `selected` field via `Rc<RefCell<...>>` so the device-change closure reads the real selection. Also fixed the fallback path: when a selected device disappears, the handler now writes the new selection directly to the shared `Rc` (previously this went through `on_selected` callbacks that were never wired up, making the fallback dead code).

### 4. `VideoEncoderConfig::new` had width and height swapped (from #552)

`web_sys::VideoEncoderConfig::new` takes `(codec, height, width)` — alphabetical order. PR #552 swapped them to `(codec, width, height)`, which would distort any non-square video. Reverted to the correct order, consistent with `screen_encoder.rs`.

### 5. Camera errors were silently swallowed

All errors in `CameraEncoder::start()` (DOM element not found, `getUserMedia` rejected, `VideoEncoder` creation failed) were logged but never reported to the user. The encoder's `enabled` flag stayed `true` even though nothing was encoding.

Added an `on_error: Callback<String>` to `CameraEncoder`, wired through `Host` → `Attendants`, matching the existing microphone error pipeline. Camera failures now show the same error dialog as microphone failures.

### 6. Dead `destroy` field removed from `EncoderState`

The `destroy` field was checked in all three encoder loops but never set to `true` anywhere in the codebase. Removed it from `EncoderState` and all encoder destructuring/loop checks to eliminate dead code and confusion.

### DIFF
--- a/videocall-client/src/lib.rs
+++ b/videocall-client/src/lib.rs
@@ -97,7 +97,8 @@
 //!     client.clone(),
 //!     "video-element",
 //!     1000000, // 1 Mbps initial bitrate
-//!     Callback::noop()
+//!     Callback::noop(),
+//!     Callback::noop() // on_error callback for camera errors
 //! );
 //! let mut microphone = create_microphone_encoder(
 //!     client.clone(),


videocall-client/src/encode/camera_encoder.rs
```diff
@@ -236,33 +238,121 @@ impl CameraEncoder {
...
+                    }
+                    return;
+                }
+            };
             let constraints = MediaStreamConstraints::new();
             let media_info = web_sys::MediaTrackConstraints::new();
-            media_info.set_device_id(&device_id.into());
+
+            // Force exact deviceId match (avoids partial/ideal matching surprises).
+            let exact = js_sys::Object::new();
```
darioalessandro, line 298: @palina-piliushyna this was the real fix!!! btw mic was broken too!! 

I am sure that chrome broke this in one of the updates!! thanks for fixing it!!

darioalessandro:
@testradav I consolidated the fixes, please take it for a spin, make sure that:
1. mic selection
2. cam selection
3. speakers selection

works every single time locally.

darioalessandro:
@palina-piliushyna you would have to accept the github invite to be a reviewer of this PR

Issue #412, opened 2025-08-23
If two phones join with the same name, they can't see each other


darioalessandro:

testradav:
some points, questions,
- is this only with phones (mobile) or also on web?  (I think on web also)
- should use the user id (not the name) to check attendees
- should allow for multiple instances of the same user id in a meeting (eg. I should be able to join the same meeting from multiple instances of the meeting with the same id)
  - each instance should have a unique id that identifies the instance separate from the user id

darioalessandro:
@anhelinakrutaya-ops I propose that before jumping to implementation we agree on the approach using this issue, please create a light design document then lets work through it. 

Do consider the discussion in https://github.com/security-union/videocall-rs/pull/529 as input to the design. 

also consider 

Requirements 
1 | Server generates session_id on connection
2 | Server sends session_id to client (new message or first RTT response)
3 | Client includes session_id in all outgoing MediaPackets
4 | Server tracks Testing → Active state, broadcasts to NATS only on Active
5 | Client delays HEARTBEAT until after election
6 | Receiving clients use session_id as peer key, email as display name
7 | Add session_id field to MediaPacket protobuf

testradav:
assigned https://github.com/anhelinakrutaya-ops

PR #567, opened 2026-02-09
If two phones join with the same name, they can't see each other #412

anhelinakrutaya-ops:
Changes:
• Add session_id field to MeetingPacket and PacketWrapper protobuf
• Server send session_id to client in MEETING_STARTED packet
• Client stores received session_id and includes it in outgoing packets
• Server tracks ConnectionState, broadcasts to NATS only when Active
• Client delays HEARTBEAT until after election completes
• Receiving clients use session_id as peer key, email as display name

### DIFF
--- a/actix-api/src/webtransport/mod.rs
+++ b/actix-api/src/webtransport/mod.rs
@@ -738,19 +738,24 @@ mod tests {
         .expect("Should receive packet from A on B");
 
         println!("Packet successfully relayed!");
-        
+
         // Parse both packets to compare content (server may add session_id)
-        let sent_packet = VcPacketWrapper::parse_from_bytes(&bytes)
-            .expect("parse sent packet");
-        let received_packet = VcPacketWrapper::parse_from_bytes(&received)
-            .expect("parse received packet");
-        
+        let sent_packet = VcPacketWrapper::parse_from_bytes(&bytes).expect("parse sent packet");
+        let received_packet =
+            VcPacketWrapper::parse_from_bytes(&received).expect("parse received packet");
+
         // Compare meaningful fields (server adds session_id, so we ignore it)
-        assert_eq!(sent_packet.packet_type, received_packet.packet_type, "packet_type must match");
+        assert_eq!(
+            sent_packet.packet_type, received_packet.packet_type,
+            "packet_type must match"
+        );
         assert_eq!(sent_packet.email, received_packet.email, "email must match");
         assert_eq!(sent_packet.data, received_packet.data, "data must match");
         // Verify that server added session_id (it should not be empty)
-        assert!(!received_packet.session_id.is_empty(), "server should add session_id");
+        assert!(
+            !received_packet.session_id.is_empty(),
+            "server should add session_id"
+        );
 
         println!("=== INTEGRATION TEST PASSED ===");
         Ok(())


darioalessandro (changes requested):
The core idea is sound: using `session_id` as a unique per-connection identifier and keying peers on it instead of `email` is the right approach to fix #412. The protobuf changes, client-side peer keying, and display-layer separation are all reasonable.

However, there are architectural and correctness issues that need to be addressed before this can ship safely.

---

### P0 — Server-Side Timer Guessing Client State

In both `ws_chat_session.rs` and `wt_chat_session.rs`, the PR introduces a hardcoded 5-second timer to "auto-activate" a connection:

```rust
ctx.run_later(std::time::Duration::from_secs(5), move |_act, _ctx| {
    let _ = addr.try_send(ActivateConnection { session: session_id.clone() });
});
```

Meanwhile, in `chat_server.rs`, all packets are silently dropped while the connection is in `Testing` state.

**Problems:**

1. **The 5-second magic number is ungrounded.** The client's `election_period_ms` is configurable and may be shorter or longer. If election takes 2s, media is suppressed for 3s unnecessarily. If election takes 7s, non-elected connections publish to NATS for 2s before cleanup — potentially causing the very duplicate-identity problem this PR aims to fix.
2. **The server is guessing at client-side state it doesn't own.** The server has no knowledge of the client's election process. This is a temporal coupling antipattern.
3. **`try_send` silently drops the activation message.** If the ChatServer actor's mailbox is full, `ActivateConnection` is silently lost, leaving the connection permanently in `Testing` state. Media would be dropped forever for that session.
4. **The timer runs per-transport-session, not per-client.** Each probe connection gets its own 5-second timer, so all connections auto-activate regardless of which one the client actually elected.

**Proposed fix — `ConnectionPhase` field on `PacketWrapper`:**

Add an enum and field to the existing `PacketWrapper` proto:

```protobuf
enum ConnectionPhase {
    CONNECTION_PHASE_UNSPECIFIED = 0;  // Old clients → treat as active (backward compat)
    PROBING = 1;
    ACTIVE = 2;
}
ConnectionPhase connection_phase = 5;
```

- Client sets `PROBING` on RTT packets during election.
- Client sets `ACTIVE` on the first heartbeat after election.
- Server transport actors read the field on inbound packets. On first `ACTIVE`, send `ActivateConnection` to `ChatServer` (using `do_send`, not `try_send`).
- **Remove the 5-second timer entirely** from both `ws_chat_session.rs` and `wt_chat_session.rs`.

This is backward compatible: old clients send `UNSPECIFIED` (proto3 default = 0), which the server treats as active — zero behavior change for existing deployments. No new packet types, no new actor messages. One enum, one field.

---

### P1 — `session_id` Never Propagated to Outgoing Heartbeats

The PR adds `session_id: Rc<RefCell<Option<String>>>` to `Connection` and reads it in the heartbeat closure:

```rust
if let Some(sid) = session_id.borrow().as_ref() {
    packet_wrapper.session_id = sid.clone();
}
```

However, **there is no code in the PR that ever writes to this `Rc<RefCell>`**. The `own_session_id` received from the `MEETING_STARTED` packet is stored on `VideoCallClient`, but never propagated to the `Connection`'s `session_id` field. Outgoing heartbeat packets will always have an empty `session_id`.

**Fix:** After receiving `MEETING_STARTED`, write the session_id into the `Connection`'s `Rc<RefCell<Option<String>>>`.

---

### P1 — Self-Packet Rejection Removed Without Replacement

The old code in `connection_manager.rs` filtered self-packets:

```rust
if packet.email != userid {
    on_inbound_media.emit(packet);
} else {
    debug!("Rejecting packet from same user: {}", packet.email);
}
```

The PR removes this check entirely and forwards all packets via `on_inbound_media.emit(packet)`. There is no replacement check using `session_id`. After the server auto-activates all connections, the client's own packets could round-trip back through NATS and create a self-peer.

The `SameUserPacket` error variant still exists in `PeerDecodeError` but is now unreachable — the triggering check was removed.

**Fix:** Restore self-packet filtering using `own_session_id` — skip packets where `packet.session_id == own_session_id`.

---

### P2 — Code Quality

1. **Inconsistent indentation** in `wt_chat_session.rs` — the `let addr` line uses different indent level (tab/space mismatch).
2. **Formatting**: `email:String` (missing space) in `peer_decode_manager.rs` Peer struct. Please run `cargo fmt`.
3. **Stale whitespace**: Extra blank lines in `peer_decode_manager.rs` decode function.
4. **`get_peer_email` swallows borrow errors**: `try_borrow()` returning `None` on failure silently masks bugs — at minimum add a `warn!` log.
5. **`reset()` passes wrong argument**: `self.video_canvas_id` instead of `self.session_id` to `new_decoders`, inconsistent with the constructor.

---

### P3 — Missing Test Coverage

Please add tests at the `actix-api` level to verify the fix works end-to-end:

- **Two clients with the same email** join a room and verify they get unique `session_id` values and can coexist.
- **`ConnectionState` transitions**: a connection in `Testing` state does not publish to NATS; after receiving a packet with `connection_phase = ACTIVE`, it does.
- **`ActivateConnection` handler** in `ChatServer`: verify idempotent `Testing → Active` transition.
- **Self-packet rejection**: client does not create a peer entry for its own `session_id`.

---

### Summary

| Priority | Issue | Action |
|----------|-------|--------|
| **P0** | Server timer guessing client state | Add `ConnectionPhase` field to proto; client signals explicitly; remove timer |
| **P1** | `session_id` never set on Connection heartbeat | Wire `own_session_id` through to Connection after MEETING_STARTED |
| **P1** | Self-packet rejection removed | Add session_id-based self-filtering |
| **P2** | Code formatting / quality | Run `cargo fmt`, fix indentation, handle borrow errors |
| **P3** | Missing tests | Add actix-api level tests for core scenarios |

The `session_id` concept is the right fix for #412. The changes above make it production-ready.

darioalessandro:
One more thing, if we change session id to be a u64 or u32 we can save a ton of transfer data, UUIDS are 36 chars 💀💀

anhelinakrutaya-ops:
> One more thing, if we change session id to be a u64 or u32 we can save a ton of transfer data, UUIDS are 36 chars 💀💀

I think it's better to create a separate task for these changes, as it's an optimization, if you don't mind. Because it also includes changes in different places, and it will probably be easier to do it after that

darioalessandro:
Hi @anhelinakrutaya-ops : if it was a UI change or something trivial I wouldn't hesitate to get it merged, but since you are introducing a protocol change, I would like to request you to change the session type within this PR, in my experience protocol changes are hard to revert/change especially in protobuf, where you can add fields to a struct but you can't change their types without breaking the schema. 

take your time.

Thanks

darioalessandro:
Taking contribution and addressing lingering concerns in merge brah `avoid-dup-nicknames`

PR #607, opened 2026-02-12
feat: Add generic OAuth2/OIDC support for meeting-api

jboyd01:
## Summary

- Replace Google-only OAuth with provider-agnostic OIDC implementation supporting Google, Okta, and any compliant provider
- Add OIDC Discovery, JWKS caching, ID token signature verification (RS256), nonce support, optional client_secret for public clients, configurable scopes, and UserInfo endpoint fallback
- Fix open redirect, CSRF state replay, and production unwrap() panics identified during security review
- Update Helm charts, Docker Compose, and documentation for the new OAuth env vars

### Backend (`meeting-api/`)
- **`oauth.rs`** — OIDC discovery, JWKS cache with rate-limited refresh, `verify_and_decode_id_token()` with sig/exp/aud/iss/nonce validation, UserInfo fallback, warning log when unverified fallback is used
- **`config.rs`** — `OAuthConfig` restructured: optional `client_secret`, `issuer`, `jwks_url`, `userinfo_url`, `scopes`; async `resolve_discovery()` for OIDC endpoint auto-population
- **`routes/oauth.rs`** — Nonce generation/validation, `return_to` open-redirect prevention (relative paths only), `unwrap()` replaced with error propagation
- **`db/oauth.rs`** — Nonce column in `oauth_requests`, CSRF state consumed atomically via `DELETE...RETURNING`
- **`state.rs`** — `JwksCache` added to `AppState`
- **`main.rs`** — Calls `resolve_discovery()` at startup
- 45 unit tests (15 new OAuth tests using local RSA keypair)

### Frontend (`yew-ui/`)
- **`login.rs`** — Replace Google-branded image button with generic "Sign in" button

image

### Infrastructure
- **`dbmate/db/migrations/20260211000000_add_oauth_nonce.sql`** — New migration
- **`docker/docker-compose.yaml`** — Updated env vars, removed Google defaults
- **`helm/meeting-api/values.yaml`** — All new OAuth env vars
- **`helm/global/us-east/meeting-api/values.yaml`** — Renamed secrets, optional secretKeyRef
- **`helm/README.md`** — Secret creation instructions
- **`Dockerfile.meeting-api`** — Runs dbmate migrations before startup
- **`cut_build_push_meeting_api.sh`** — New build script

### Documentation
- **`docs/OAUTH_HELM_CONFIGURATION.md`** — Provider-specific setup guide (Google, Okta, generic OIDC)
- **`docs/MEETING_OWNERSHIP.md`** — Updated env var table, provider-agnostic language

## Security review findings addressed
1. **Open redirect** (HIGH) — `return_to` validated to relative paths only
2. **CSRF state replay** (HIGH) — State token consumed atomically on use
3. **Production panics** (MEDIUM) — `unwrap()` on `HeaderValue::from_str()` replaced with `map_err`
4. **Unverified fallback** (HIGH) — Warning logged when JWKS not configured

## Test plan
- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy -p meeting-api -- -D warnings` passes
- [x] 45/45 unit tests pass (`cargo test -p meeting-api --lib`)
- [x] Integration tests pass in Docker (`cargo test -p meeting-api` with `DATABASE_URL`)
- [x] Deploy with Google provider — verify OIDC discovery, login, callback, session cookie
- [x] Deploy with Okta provider (public client, no secret) — verify PKCE-only flow
- [x] Verify `return_to` rejects absolute URLs
- [x] Verify CSRF state cannot be replayed (second callback with same state returns error)

### DIFF
--- a/.release-plz.toml
+++ b/.release-plz.toml
@@ -33,6 +33,12 @@ changelog_update = true
 git_release_enable = true
 
 [[package]]
+name = "videocall-meeting-client"
+publish = true
+changelog_update = true
+git_release_enable = true
+
+[[package]]
 name = "bot"
 publish = true
 changelog_update = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2507,26 +2507,6 @@ dependencies = [
 
 [[package]]
 name = "gloo-net"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2899cb1a13be9020b010967adc6b2a8a343b6f1428b90238c9d53ca24decc6db"
-dependencies = [
- "futures-channel",
- "futures-core",
- "futures-sink",
- "gloo-utils 0.1.7",
- "js-sys",
- "pin-project",
- "serde",
- "serde_json",
- "thiserror 1.0.69",
- "wasm-bindgen",
- "wasm-bindgen-futures",
- "web-sys",
-]
-
-[[package]]
-name = "gloo-net"
 version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9902a044653b26b99f7e3693a42f171312d9be8b26b5697bd1e43ad1f8a35e10"
@@ -5212,15 +5192,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
-name = "reqwasm"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05b89870d729c501fa7a68c43bf4d938bbb3a8c156d333d90faa0e8b3e3212fb"
-dependencies = [
- "gloo-net 0.1.0",
-]
-
-[[package]]
 name = "reqwest"
 version = "0.11.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -7345,6 +7316,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "videocall-meeting-client"
+version = "0.1.0"
+dependencies = [
+ "log",
+ "reqwest 0.12.26",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.17",
+ "videocall-meeting-types",
+]
+
+[[package]]
 name = "videocall-meeting-types"
 version = "0.1.0"
 dependencies = [
@@ -7475,7 +7458,6 @@ dependencies = [
  "neteq",
  "once_cell",
  "regex",
- "reqwasm",
  "reqwest 0.12.26",
  "serde",
  "serde-wasm-bindgen 0.6.5",
@@ -7483,6 +7465,8 @@ dependencies = [
  "urlencoding",
  "videocall-client",
  "videocall-diagnostics",
+ "videocall-meeting-client",
+ "videocall-meeting-types",
  "videocall-types",
  "wasm-bindgen",
  "wasm-bindgen-futures",
@@ -7542,8 +7526,6 @@ dependencies = [
  "cfg-if 1.0.4",
  "once_cell",
  "rustversion",
- "serde",
- "serde_json",
  "wasm-bindgen-macro",
  "wasm-bindgen-shared",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "bot",
     "videocall-types",
     "videocall-meeting-types",
+    "videocall-meeting-client",
     "videocall-client",
     "videocall-cli",
     "yew-ui",
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ An open-source, high-performance video conferencing platform built with Rust, pr
   - [Local/Docker](#localdocker-start-yewsh)
   - [Kubernetes/Helm](#kuberneteshelm-configmap-configjsyaml)
 - [Usage](#usage)
+- [Meeting Management](#meeting-management)
 - [Performance](#performance)
 - [Security](#security)
 - [Feature Flags](#feature-flags)
@@ -242,6 +243,45 @@ videocall-cli stream \
 ```
 
 For detailed information about the CLI tool and all available options, see the [videocall-cli README](videocall-cli/README.md).
+
+## Meeting Management
+
+videocall.rs includes a comprehensive meeting management system with ownership, waiting rooms, and host controls.
+
+### Key Features
+
+- **Meeting Ownership**: Each meeting has an owner (the creator) identified by their email
+- **My Meetings**: Users can view and manage all meetings they own from the home page
+- **Waiting Room**: Non-owners enter a waiting room and must be admitted by an existing participant
+- **Host Identification**: The meeting owner is visually identified with "(Host)" in the UI
+- **Soft Delete**: Owners can delete their meetings; deleted meeting IDs can be reused
+
+### Meeting Workflow
+
+1. **Create/Join**: Navigate to `/meeting/{meeting-id}` - if the meeting doesn't exist, you become the owner
+2. **Start/Join Button**: Owners see "Start Meeting", others see "Join Meeting"
+3. **Waiting Room**: Non-owners wait for admission; admitted participants can manage the waiting room
+4. **Auto-Join**: When admitted from the waiting room, participants automatically enter the meeting
+
+### Documentation
+
+For detailed information about the meeting system:
+
+- **[Meeting Ownership & Workflow](docs/MEETING_OWNERSHIP.md)** - Ownership model, lifecycle, and user workflows
+- **[Meeting API Reference](docs/MEETING_API.md)** - Complete API endpoint documentation
+
+### Enabling Meeting Management
+
+Meeting management requires the `FEATURE_MEETING_MANAGEMENT` flag:
+
+```bash
+export FEATURE_MEETING_MANAGEMENT=true
+```
+
+Or in Docker:
+```bash
+docker run -e FEATURE_MEETING_MANAGEMENT=true ...
+```
 
 ## Performance
 
--- a/actix-api/src/bin/websocket_server.rs
+++ b/actix-api/src/bin/websocket_server.rs
@@ -28,7 +28,6 @@ use actix_web::{
 use reqwest::header::LOCATION;
 use sec_api::{
     actors::chat_server::ChatServer,
-    api,
     auth::{
         fetch_oauth_request, generate_and_store_oauth_request, request_token, upsert_user,
         AuthRequest,
@@ -328,7 +327,6 @@ async fn main() -> std::io::Result<()> {
                 .service(logout)
                 .service(ws_connect_authenticated)
                 .service(ws_connect)
-                .configure(api::configure_api_routes)
         } else if db_enabled {
             // OAuth requires database (r2d2 pool for legacy OAuth code)
             let pool = get_pool();
@@ -356,7 +354,6 @@ async fn main() -> std::io::Result<()> {
                 .service(logout)
                 .service(ws_connect_authenticated)
                 .service(ws_connect)
-                .configure(api::configure_api_routes)
         } else {
             // OAuth configured but database disabled - skip OAuth routes
             error!("OAuth is configured but DATABASE_ENABLED=false. OAuth requires database. Skipping OAuth routes.");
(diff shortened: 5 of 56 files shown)


helm/global/us-east/meeting-api/values.yaml
```diff
@@ -37,18 +37,26 @@ meeting-api:
...
         secretKeyRef:
-          name: google-oauth-credentials
+          name: oauth-credentials
           key: client-id
+          optional: true
     - name: OAUTH_SECRET
       valueFrom:
         secretKeyRef:
-          name: google-oauth-credentials
+          name: oauth-credentials
```
jboyd01, line 50: Note the cookie name change!

darioalessandro:
Hey Jay, I do want to keep the Google button for people using Gmail and to show an okra button if okta auth is enabled you could drive that with ui config Sent from my iPhoneOn Feb 12, 2026, at 14:02, Jay Boyd ***@***.***> wrote:﻿

@jboyd01 requested your review on: security-union/videocall-rs#607 feat: Add generic OAuth2/OIDC support for meeting-api.

darioalessandro (commented):
.

meeting-api/src/oauth.rs
```diff
@@ -11,24 +11,326 @@
...
+    #[serde(default)]
+    pub family_name: Option<String>,
+    #[serde(default)]
+    pub nonce: Option<String>,
+    /// Standard OIDC `iss` claim.
+    #[serde(default)]
+    pub iss: Option<String>,
+    /// Standard OIDC `aud` — can be a single string or array.
+    /// We handle both forms during validation; this captures the raw value.
+    #[serde(default)]
```
darioalessandro: `prompt=select_account` is Google-specific. Other providers might choke on it. Make it configurable or drop it.

meeting-api/src/oauth.rs
```diff
@@ -11,24 +11,326 @@
...
+    let url = format!(
+        "{}/.well-known/openid-configuration",
+        issuer.trim_end_matches('/')
+    );
+    let resp = reqwest::get(&url)
+        .await
+        .map_err(|e| AppError::internal(&format!("OIDC discovery request failed: {e}")))?;
+
+    if !resp.status().is_success() {
+        let status = resp.status();
```
darioalessandro, line 101: No `unsafe` needed here. Just do `let mut bytes = token.into_bytes(); bytes[len-1] = ...; let token = String::from_utf8(bytes).unwrap();`

cut_build_push_meeting_api.sh
```diff
@@ -0,0 +1,18 @@
...
+TAG=$1
+if [ -z "$1" ]
+then
+    TAG=$(git rev-parse HEAD)
+fi
+
+IMAGE_URL=securityunion/videocall-meeting-api:$TAG
+echo "Building image "$IMAGE_URL
+
+if ! docker build -t $IMAGE_URL . --file Dockerfile.meeting-api; then
```
darioalessandro, line 13: Quote `"$IMAGE_URL"` — if the tag has special chars this breaks.

meeting-api/src/oauth.rs
darioalessandro, line 54: `prompt=select_account` is Google-specific. Okta and other providers might reject this. Make it a per-provider config field.

meeting-api/src/oauth.rs
```diff
@@ -11,24 +11,326 @@
...
     pub name: String,
+    #[serde(default)]
+    pub email_verified: Option<bool>,
+    #[serde(default)]
+    pub given_name: Option<String>,
+    #[serde(default)]
+    pub family_name: Option<String>,
+    #[serde(default)]
+    pub nonce: Option<String>,
+    /// Standard OIDC `iss` claim.
```
darioalessandro, line 49: None of these params are URL-encoded. `scopes` has literal spaces in the query string. Technically broken per RFC 3986 -- use `urlencoding::encode()`.

meeting-api/src/oauth.rs
darioalessandro, line 101: No `unsafe` needed here. Just do `let mut bytes = token.into_bytes(); bytes[len-1] = ...; let token = String::from_utf8(bytes).unwrap();`

cut_build_push_meeting_api.sh
darioalessandro, line 13: Quote `"$IMAGE_URL"` -- if the tag has special chars this breaks.

darioalessandro (commented):
Hey @jboyd01 -- great work on this. The security fixes are solid, especially the nonce addition for OIDC replay prevention and the `DELETE...RETURNING` for atomic CSRF state consumption. Those were real vulnerabilities and you knocked them out clean.

A few things before we merge:

**1. `oauth.rs` is 851 lines.** Our rule is 200-300 max per file. This has 6 concerns crammed into one module (claims, discovery, JWKS cache, JWT verification, token exchange, UserInfo). Split it into `oauth/` submodules.

**2. `prompt=select_account` is hardcoded** in `build_auth_url`. That's Google-specific -- other providers may reject it. Make it configurable per provider (or remove it).

**3. `access_type=offline` was removed.** The old code sent this to Google to get refresh tokens. Since we still store refresh tokens via `upsert_user`, this is a regression for Google deployments. Needs to be configurable.

**4. URL params in `build_auth_url` aren't URL-encoded.** `scopes` has literal spaces in the query string. Use `urlencoding::encode()` or the `url` crate.

**5. `unsafe` in the tampered token test is unnecessary.** Use `token.into_bytes()` / mutate / `String::from_utf8()` instead.

**6. Branding: Per Google OAuth guidelines we need to show a "Login with Google" button when we use them so we can't make the Login button completely generic: https://developers.google.com/identity/branding-guidelines else they do not let you get your app certified, probably Okta is similar

### Design: multi-provider support

This PR sets up a single-provider architecture, but in practice most apps need multiple providers on the same login page ("Sign in with Google" + "Sign in with Okta" + "Sign in with GitHub"). I'd like us to design for that from the start so we don't have to rip things apart later. Here's my proposal:

**Shared `OAuthProvider` enum in `videocall-meeting-types`:** Define a `OAuthProvider` enum in the shared types crate so that provider IDs are type-safe and consistent across backend and frontend:

```rust
// videocall-meeting-types/src/oauth.rs
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
#[serde(rename_all = "snake_case")]
pub enum OAuthProvider {
    Google,
    Okta,
    GitHub,
    Microsoft,
    Custom, // for arbitrary OIDC providers
}
```

This gives us:
- **Backend**: match on the enum to apply provider-specific behavior (default scopes, prompt, extra params). The config file maps to this enum.
- **Frontend**: match on the enum to pick the right logo, button style, and display name. No string matching, no typos.
- **Serde**: serializes to `"google"`, `"okta"`, etc. in JSON -- clean for both the config file and the API.

Add `videocall-meeting-types` as a dependency of `yew-ui` (it's not one today) so both sides share the enum.

**Config:** Instead of flat env vars for one provider, support a JSON config file (`OAUTH_PROVIDERS_CONFIG` env var points to the path). Each entry uses the `OAuthProvider` enum as its `provider` field, plus `display_name`, `issuer`, `scopes`, `prompt`, `extra_auth_params`, etc. Secrets come from env vars by convention: `OAUTH_{PROVIDER}_CLIENT_ID`, `OAUTH_{PROVIDER}_CLIENT_SECRET`. For backward compat, if `OAUTH_PROVIDERS_CONFIG` is not set, fall back to the existing single-provider env vars with provider = `Google`.

Example config:
```json
[
  {
    "provider": "google",
    "display_name": "Sign in with Google",
    "issuer": "https://accounts.google.com",
    "scopes": "openid email profile",
    "prompt": "select_account",
    "extra_auth_params": { "access_type": "offline" }
  },
  {
    "provider": "okta",
    "display_name": "Sign in with Okta",
    "issuer": "https://acme.okta.com",
    "scopes": "openid email profile"
  }
]
```

**Backend:**
- `AppState` holds a `HashMap<OAuthProvider, ProviderConfig>` instead of `Option<OAuthConfig>`, plus per-provider JWKS caches.
- Routes: `GET /login/{provider}` initiates the flow for that provider. `GET /login/callback` stays the same URL -- provider is resolved from the stored `oauth_request` (add a `provider` column to `oauth_requests`).
- `GET /login` with no provider redirects to `/login/{first_provider}` for backward compat.
- OIDC discovery runs per provider at startup.

**DB:** New migration adds `provider VARCHAR(64) NOT NULL DEFAULT 'google'` to `oauth_requests`. `store_oauth_request` and `fetch_oauth_request` include it.

**Frontend:** The provider list gets baked into `window.__APP_CONFIG` at deploy time (same pattern as `apiBaseUrl`, `oauthEnabled`, etc.). The login component reads the list from config and `match`es on `OAuthProvider` to pick the right logo SVG and button style -- no string comparisons. We ship built-in SVGs for the enum variants (Google, GitHub, Microsoft, Okta). Each button navigates to `{apiBaseUrl}/login/{provider}?returnTo=...`. If only one provider is configured, the UX is identical to today.

**User identity stays email-based.** Same email from different providers = same user. Session JWT is provider-agnostic.

This also solves the `prompt` and `extra_auth_params` issues from the review -- they become per-provider config fields instead of hardcoded values.

### Nits

- `cut_build_push_meeting_api.sh` -- quote `"$IMAGE_URL"` in the docker commands.
- `is_none_or()` requires Rust 1.82+ -- confirm that's our MSRV.

cut_build_push_meeting_api.sh
jboyd01: Thanks for the thorough review, @darioalessandro.  I agree on all 5 concrete issues — I'll fix them in this PR:

1. Splitting `oauth.rs` into submodules
2. Making `prompt` and `extra_auth_params` configurable (fixes both the `select_account` and `access_type=offline` regressions)
3. URL-encoding auth params with the `url` crate
4. Removing `unsafe` from the test
5. Quoting `$IMAGE_URL` in the build script
6. Revert use of Google branded login button

Regarding minimal rust version: `rust-version` is not set in any `Cargo.toml`, the dependency tree already requires a recent toolchain: axum 0.7.9 needs Rust 1.75+, reqwest 0.5.0 needs 1.80+.  So `is_none_or()` (stabilized in 1.82) is a small bump.   ~~How about I document MSRV as 1.82 and add it to `Cargo.toml` ?~~   And I will update https://github.com/security-union/videocall-rs/blob/main/README.md?plain=1#L127 which states `1.89+`.   Oh, check it out, `yew-ui/rust-toolchain.toml` pins `1.85` as does `matomo-logger/rust-toolchain.toml`.   Looks like we should declare MSRV as `1.85`.

On multi-provider: I absolutely agree that's the right direction. I'd like to propose landing this PR with the above fixes (keeping it focused on the security improvements + cleanup), then immediately following up with a multi-provider PR that adds the `OAuthProvider` enum, per-provider config, `GET /login/{provider}`, and the DB migration. Frontend branded buttons would be a third PR. That way each change is reviewable in isolation and we don't risk the security fixes getting stalled behind a larger design effort.   And I'll revert the google branded login button here in this PR.   Sound ok?  Thoughts?

jboyd01:
All 6 review items addressed:

1. **oauth.rs split** — Now oauth/ with 5 submodules (claims, discovery, jwks, verify, exchange). Public API unchanged via re-exports from mod.rs.

2. **prompt=select_account removed** — New OAUTH_PROMPT env var. Omitted from auth URL when unset.

3. **access_type=offline regression fixed** — New OAUTH_EXTRA_PARAMS env var (JSON object). Google deployments set `OAUTH_EXTRA_PARAMS={"access_type":"offline"}`. Documented in Helm values and OAUTH_HELM_CONFIGURATION.md.

4. **URL encoding** — build_auth_url() now uses url::Url with query_pairs_mut().append_pair(). New tests verify no literal spaces.

5. **unsafe removed** — Safe into_bytes() + from_utf8().unwrap() in tampered token test.

6. **$IMAGE_URL quoted** in cut_build_push_meeting_api.sh.

**MSRV:** Declared rust-version = "1.85" in meeting-api/Cargo.toml. Fixed README (1.89 → 1.85).

**Branding:** Added GET /auth/config endpoint that derives the provider name from OAUTH_ISSUER at startup (accounts.google.com → "google", *.okta.com → "okta", etc.). The frontend fetches this on the login page and renders a branded "Sign in with Google" button with the standard G logo SVG when applicable. No duplicate UI-side config needed.

**Multi-provider:** Agree on the direction — proposing as a follow-up PR to keep this one focused on the security fixes and review feedback.

helm/global/us-east/meeting-api/values.yaml
darioalessandro: 10 4

darioalessandro:
I did not see the generic UI page so I am adding this here including your changes @jboyd01

PR #610, opened 2026-02-15
add generic oauth support

darioalessandro:
## Summary

Continuation of #607 by @jboyd01. Replaces the Google-only OAuth login with a provider-agnostic OIDC implementation that supports Google, Okta, and any compliant provider.

### Why

The original implementation was hardwired to Google: hardcoded endpoints, Google-branded button, no token signature verification. This made it impossible to use any other identity provider and left the ID token unverified (trusted purely because it came over HTTPS).

### What changed

**Security fixes:**
- **CSRF state replay prevention** -- `fetch_oauth_request` changed from `SELECT` to `DELETE...RETURNING` so each state token can only be used once
- **Open redirect prevention** -- `return_to` parameter validated to relative paths only (rejects absolute URLs and `//` protocol-relative)
- **Nonce support** -- OIDC nonce generated at login, stored in DB, validated on the ID token to prevent replay attacks
- **ID token signature verification** -- JWKS-based RS256 verification with cached keys and rate-limited refresh (5 min)
- **Production panic removal** -- `unwrap()` on `HeaderValue::from_str()` replaced with proper error propagation in callback and logout

**OIDC support:**
- OIDC Discovery: set `OAUTH_ISSUER` and endpoints are auto-populated from `.well-known/openid-configuration`
- Optional `client_secret` for public clients (PKCE-only flows like Okta SPA)
- Configurable scopes via `OAUTH_SCOPES`
- UserInfo endpoint fallback when ID token lacks `email` claim
- `email` field is now `Option<String>` to correctly model providers that don't include it

**Frontend:**
- Provider-branded login button: reads `oauthProvider` from `window.__APP_CONFIG` and renders the matching logo (Google "G" SVG, Okta logo) and display text
- Falls back to generic "Sign in" when no provider is specified

**Infrastructure:**
- New DB migration: `nonce` column on `oauth_requests`
- Helm values and Docker Compose updated for new env vars (`OAUTH_ISSUER`, `OAUTH_JWKS_URL`, `OAUTH_USERINFO_URL`, `OAUTH_SCOPES`, `OAUTH_PROVIDER`)
- Dockerfile runs dbmate migrations before startup
- Provider-specific setup docs for Google, Okta, and generic OIDC

### How to test

1. `cargo fmt --all -- --check`
2. `cargo clippy -p meeting-api -- -D warnings`
3. `cargo test -p meeting-api --lib` (unit tests including 15 new JWT verification tests with real RSA keypairs)
4. Deploy with Google: set `OAUTH_ISSUER=https://accounts.google.com` and verify OIDC discovery, login, callback, session cookie
5. Deploy with Okta (public client, no secret): verify PKCE-only flow
6. Verify `return_to` rejects absolute URLs
7. Verify CSRF state cannot be replayed (second callback with same state returns error)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7244,7 +7244,7 @@ dependencies = [
 
 [[package]]
 name = "videocall-client"
-version = "3.0.0"
+version = "4.0.0"
 dependencies = [
  "aes",
  "anyhow",
@@ -7281,7 +7281,7 @@ dependencies = [
 
 [[package]]
 name = "videocall-codecs"
-version = "0.1.11"
+version = "0.1.12"
 dependencies = [
  "anyhow",
  "approx",
@@ -7329,7 +7329,7 @@ dependencies = [
 
 [[package]]
 name = "videocall-meeting-types"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "serde",
 ]
@@ -7408,7 +7408,7 @@ dependencies = [
 
 [[package]]
 name = "videocall-sdk"
-version = "0.1.13"
+version = "0.1.14"
 dependencies = [
  "anyhow",
  "bytes",
@@ -7443,7 +7443,7 @@ dependencies = [
 
 [[package]]
 name = "videocall-ui"
-version = "1.1.35"
+version = "1.1.36"
 dependencies = [
  "anyhow",
  "console_error_panic_hook",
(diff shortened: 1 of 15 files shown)


darioalessandro:
@cursoragent please review, especial emphasis on not breaking Google OAuth

yew-ui/tests/login_provider_logo.rs
```diff
@@ -0,0 +1,201 @@
+// Copyright 2025 Security Union LLC
```
darioalessandro, line 1: @jboyd01 now we can do super clean UI tests like this

PR #612, opened 2026-02-15
Tweaks to deploy OAuth and meeting API

darioalessandro:


helm/meeting-api/templates/deployment.yaml
```diff
@@ -32,7 +32,6 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          command: ['meeting-api']
```
darioalessandro, line 35: heads up @jboyd01 this allows the meeting-api container to run the default command 

https://github.com/security-union/videocall-rs/blob/fee3b236fac3e0cfbdc801a9ac2dbbda8bc3f671/Dockerfile.meeting-api#L18

so that db migrations can run, this is important because the streaming-server no longer runs migrations.

PR #649, opened 2026-02-22
Fix: Same-display-name peers can't see each other (Closes #412)

### LINKED ISSUE #412: If two phones join with the same name, they can't see each other

darioalessandro:
Introduced a `u64` session id that the system uses to keep track of peers, now the username is not a factor.

### DIFF
--- a/dioxus-ui/src/components/attendants.rs
+++ b/dioxus-ui/src/components/attendants.rs
@@ -128,7 +128,7 @@ pub fn AttendantsComponent(
     let mut show_dropdown = use_signal(|| false);
     let meeting_ended_message = use_signal(|| None::<String>);
     let mut meeting_info_open = use_signal(|| false);
-    let peers = use_signal(Vec::<String>::new);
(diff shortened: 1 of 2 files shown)


yew-ui/src/pages/meeting.rs
```diff
@@ -530,9 +530,13 @@ pub fn meeting_page(props: &MeetingPageProps) -> Html {
...
                     // User is admitted - show the meeting
                     (Some(username), MeetingStatus::Admitted { is_host, host_display_name, room_token }) => {
+                        // Use canonical email from meeting API for media connection identity (must match JWT sub).
+                        // Display name (username) is for UI only.
+                        let connection_email = (*current_user_email).clone()
+                            .unwrap_or_else(|| username.clone());
                         html! {
                             <AttendantsComponent
-                                email={username.clone()}
+                                email={connection_email}
```
darioalessandro: revert

PR #668, opened 2026-02-24
Revert NIX backend images from prod

darioalessandro:
CI was producing broken Images

### DIFF
--- a/Dockerfile.meeting-api
+++ b/Dockerfile.meeting-api
@@ -1,37 +1,18 @@
-# Stage 1: Nix build with musl target
-FROM nixos/nix:2.33.2 AS builder
-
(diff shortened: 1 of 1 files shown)


Dockerfile.actix
```diff
@@ -1,38 +1,25 @@
-FROM nixos/nix:2.33.2 AS builder
```
darioalessandro, line 1: @jboyd01 walk of shame: reverting nix changes here until I get it right XD

jboyd01: LOL, I'm glad I'm not the only one on that road at times!  I spent some time trying to get it working only to give up.

testradav: This begs for a full test suite :)

PR #675, opened 2026-02-26
fix: make session cookie name configurable to prevent PR preview 401 collision

jboyd01:
@darioalessandro  this fixes the issue we saw with cookie collision in PR Previews - presently PR Previews best work in incognito to avoid cookie name collision.

## Summary

- Production `api.videocall.rs` sets a `session` cookie with `Domain=.videocall.rs`. PR preview APIs at `pr{N}-api.sandbox.videocall.rs` are subdomains of `videocall.rs`, so browsers send the production cookie there too. The preview API validates it with a different `JWT_SECRET` → 401.
- Confirmed by: auth works in incognito (no production cookies), fails in a normal browser session.
- Fix: add `COOKIE_NAME` env var (default: `"session"`). PR previews set `COOKIE_NAME=pr${SLOT}-session` (e.g. `pr1-session`, `pr2-session`) — distinct from production and from each other.

## Changes

- `meeting-api/src/config.rs` — add `cookie_name` field, read from `COOKIE_NAME` env var
- `meeting-api/src/state.rs` — propagate `cookie_name` through `AppState`
- `meeting-api/src/routes/oauth.rs` — use `state.cookie_name` in `Set-Cookie` headers
- `meeting-api/src/auth.rs` — use `state.cookie_name` when parsing `Cookie` header
- `helm/meeting-api/values.yaml` — document `COOKIE_NAME` with default `"session"`
- `.github/workflows/pr-deploy-reusable.yaml` — set `COOKIE_NAME=pr${SLOT}-session` for preview deployments

## Test plan

- [ ] 66 unit tests pass (`cargo test -p meeting-api`)
- [ ] New regression test `production_session_cookie_rejected_by_preview_api` explicitly guards against the collision bug
- [ ] New tests for `build_session_cookie` / `build_clear_session_cookie` cookie builders
- [ ] After merge, redeploy a PR preview and confirm OAuth login no longer returns 401

> **Note:** The workflow change in `.github/workflows/pr-deploy-reusable.yaml` must be merged to `main` before it takes effect — `issue_comment` triggered workflows always run from the default branch.

### DIFF
--- a/meeting-api/src/routes/oauth.rs
+++ b/meeting-api/src/routes/oauth.rs
@@ -40,8 +40,14 @@ use crate::token;
 // ---------------------------------------------------------------------------
 
 /// Build a `Set-Cookie` header value for the session JWT.
-fn build_session_cookie(jwt: &str, ttl_secs: i64, domain: Option<&str>, secure: bool) -> String {
-    let mut cookie = format!("session={jwt}; Path=/; HttpOnly; SameSite=Lax; Max-Age={ttl_secs}");
+fn build_session_cookie(
+    name: &str,
+    jwt: &str,
+    ttl_secs: i64,
+    domain: Option<&str>,
+    secure: bool,
+) -> String {
+    let mut cookie = format!("{name}={jwt}; Path=/; HttpOnly; SameSite=Lax; Max-Age={ttl_secs}");
     if secure {
         cookie.push_str("; Secure");
     }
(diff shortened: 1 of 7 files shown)


meeting-api/src/routes/oauth.rs
```diff
@@ -281,3 +292,83 @@ pub async fn logout(State(state): State<AppState>) -> Result<Response, AppError>
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
```
darioalessandro, line 297: tests

darioalessandro:
@jboyd01 after deleting the `videocall.rs` cookie it works

Screenshot 2026-02-26 at 22 39 17

PR #682, opened 2026-02-26
fix/redirect-login-frontends:  Fix OAuth redirect to return to the correct frontend

testradav:
Fixes the return URL after authentication to the originating frontend client app


## Fix OAuth redirect to return to the correct frontend                                                                                                                                                                                             
                                                                                                     
  After OAuth login, the callback handler always redirected to `AFTER_LOGIN_URL` (yew-ui on port 80), even when the user authenticated from dioxus-ui (port 3001). Additionally, absolute `returnTo` URLs from meeting pages were silently rejected by
   the backend's relative-path-only sanitization.                                                                                                                                                 
                                                                                                                                                                                                                                                      
### Changes                                                                                        
                                         
  **Backend (`meeting-api`)**
  - Added `url` crate for origin-based URL validation
  - Added `ALLOWED_REDIRECT_URLS` config (comma-separated origins allowlist)
  - Replaced inline `returnTo` sanitization with `validate_return_to()` — accepts relative paths and absolute URLs whose origin matches `AFTER_LOGIN_URL` or the allowlist
  - Callback handler now uses absolute `returnTo` URLs directly instead of always prepending `AFTER_LOGIN_URL`
  - Added 12 unit tests covering relative paths, allowed/disallowed origins, protocol-relative, `javascript:`, scheme/port mismatches

  **Frontends (`yew-ui`, `dioxus-ui`)**
  - Login button now sends `?returnTo={origin}/` when no query string exists, so the backend knows which frontend to redirect back to

  **Config (`docker-compose`, Helm, CI)**
  - Added `ALLOWED_REDIRECT_URLS` env var to meeting-api service in docker-compose, Helm values, and PR preview workflow


## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [x] Documentation updated

## Testing
- [ ] No tests needed


see `returnTo` param on login URL
<img width="2007" height="813" alt="image" src="https://github.com/user-attachments/assets/40c95b0a-2bea-4a35-87d1-45edff5e3102" />


docker/docker-compose.yaml
```diff
@@ -128,6 +128,7 @@ services:
       - OAUTH_USERINFO_URL=${OAUTH_USERINFO_URL:-}
       - OAUTH_SCOPES=${OAUTH_SCOPES:-openid email profile}
       - AFTER_LOGIN_URL=${AFTER_LOGIN_URL:-http://localhost:${TRUNK_SERVE_PORT:-80}}
+      - ALLOWED_REDIRECT_URLS=http://localhost:${TRUNK_SERVE_PORT:-80},http://localhost:${DIOXUS_SERVE_PORT:-3001}
```
testradav: This needs to be changed do not hardcode local host before the environment variable

Issue #685, opened 2026-02-27
Users with different account IDs but the same display name are identified as Host at the same time


anhelinakrutaya-ops:
When a user with a specific account ID creates a meeting and is assigned the Host role, another user with a different account ID but the same display name is also shown as Host. Host assignment is based on display name, causing multiple users to appear as Hosts simultaneously.

PR #686, opened 2026-02-27
Users with different account IDs but the same display name are displayed as Host at the same time

anhelinakrutaya-ops:
Host display is determined by the email address than the display name #685

### DIFF
--- a/videocall-client/src/encode/camera_encoder.rs
+++ b/videocall-client/src/encode/camera_encoder.rs
@@ -188,7 +188,6 @@ impl CameraEncoder {
         // 2. setup WebCodecs, in particular
         // 3. send encoded video frames and raw audio to the server.
         let client = self.client.clone();
-        let userid = client.userid().clone();
         let aes = client.aes();
         let video_elem_id = self.video_elem_id.clone();
         let EncoderState {
@@ -226,7 +225,6 @@ impl CameraEncoder {
                     chunk,
                     sequence_number,
                     buffer.as_mut_slice(),
-                    &userid,
                     aes.clone(),
                 );
                 client.send_packet(packet);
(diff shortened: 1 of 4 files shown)


actix-api/src/actors/chat_server.rs
```diff
@@ -229,6 +229,11 @@ impl Handler<ClientMessage> for ChatServer {
                 if packet_wrapper.session_id == 0 {
                     packet_wrapper.session_id = session;
                 }
+
+                if packet_wrapper.user_email.is_empty() {
```
darioalessandro, line 233: makes sense

yew-ui/src/components/canvas_generator.rs
```diff
@@ -36,11 +36,14 @@ pub fn generate_for_peer(
...
-    host_display_name: Option<&str>,
+    host_email: Option<&str>,
 ) -> Html {
     let peer_email = client.get_peer_email(key).unwrap_or_else(|| key.clone());
+    let user_email = client
+        .get_peer_user_email(key)
+        .unwrap_or_else(|| key.clone());
 
-    let is_host = host_display_name.map(|h| h == peer_email).unwrap_or(false);
+    let is_host = host_email.map(|h| h == user_email).unwrap_or(false);
```
darioalessandro, line 46: makes sense.

protobuf/types/packet_wrapper.proto
```diff
@@ -17,4 +17,5 @@ message PacketWrapper {
   string email = 2;
```
darioalessandro, line 17: @anhelinakrutaya-ops there's already an email why do we need another email?

anhelinakrutaya-ops: @darioalessandro It's labeled as "email", but it's being used as a display name. I'm not sure if I can use it as a real email, as it might affect other parts of the code

anhelinakrutaya-ops:
Was fixed by #724

PR #688, opened 2026-02-27
fix: waiting room dioxis app

testradav:
Fix dioxus-ui guest stuck on "Waiting for meeting" and add missing reconnection after connection loss

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Documentation updated

## Testing
- [x] No tests needed

image


darioalessandro (approved):
lgtm would be nice if you could add wasm test for the reconnect code to make sure that we never break this

PR #690, opened 2026-02-28
Show peer display name on video tile and peer list

iamgabrielsoft:
This PR fixes the display of peer names in the video call interface, showing actual display names instead of raw session IDs.

### VideoCallClient Enhancement
- **Added `get_peer_display_name()` method** that reads `peer.display_name` from `PeerDecodeManager` (updated via heartbeat packets)
- Provides a clean API to access the current display name for any peer

### Canvas Generator Updates  
- **Updated floating name labels** on video tiles (full-bleed, grid, and screen-share layouts) to use `get_peer_display_name()`
- **Added `peer_display_name` prop** to `PeerTileProps` so Yew re-renders the tile when the name changes
- Props change triggers re-render, ensuring UI stays in sync with name updates

### Peer List Improvements
- **Pass `peer_display_names` map** to `PeerList` component 
- **Sidebar attendant list** now shows real names instead of raw session_id numbers
- Improves user experience by making it easier to identify participants

### Attendants Component Updates
- **Pass `peer_display_name` per-tile** in `attendants.rs` `view()` function
- Ensures each video tile displays the correct current name for that peer

### Testing Performed
- Verified display names update correctly when changed via heartbeat
- Confirmed UI re-renders with new names
- Tested both grid and full-bleed layouts show proper names
- Sidebar participant list shows real names instead of session IDs

### DIFF
--- a/videocall-client/src/client/video_call_client.rs
+++ b/videocall-client/src/client/video_call_client.rs
@@ -567,6 +567,24 @@ impl VideoCallClient {
         }
     }
 
+    /// Returns the current display name for a peer (updated via heartbeat).
+    /// Falls back to the session_id string if no display name is set yet.
+    pub fn get_peer_display_name(&self, session_id: &str) -> Option<String> {
+        let sid: u64 = session_id.parse().ok()?;
+        match self.inner.try_borrow() {
+            Ok(inner) => inner
+                .peer_decode_manager
+                .get(&sid)
+                .map(|peer| peer.display_name.clone()),
+            Err(_) => {
+                warn!(
+                    "Failed to borrow inner in get_peer_display_name for session_id: {session_id}"
+                );
+                None
+            }
+        }
+    }
+
     /// Hacky function that returns true if the given peer has yet to send a frame of screen share.
     ///
     /// No reason for this function to exist, it should be deducible from the
--- a/yew-ui/src/components/attendants.rs
+++ b/yew-ui/src/components/attendants.rs
@@ -941,7 +941,8 @@ impl Component for AttendantsComponent {
             .map(|(i, peer_id)| {
                 let full_bleed = display_peers_vec.len() == 1
                     && !self.client.is_screen_share_enabled_for_peer(peer_id);
-                html!{ <PeerTile key={format!("tile-{}-{}", i, peer_id)} peer_id={peer_id.clone()} full_bleed={full_bleed} host_display_name={host_display_name.clone()} /> }
+                let peer_display_name = self.peer_display_names.get(peer_id).cloned();
+                html!{ <PeerTile key={format!("tile-{}-{}", i, peer_id)} peer_id={peer_id.clone()} full_bleed={full_bleed} host_display_name={host_display_name.clone()} peer_display_name={peer_display_name} /> }
             })
             .collect();
 
--- a/yew-ui/src/components/canvas_generator.rs
+++ b/yew-ui/src/components/canvas_generator.rs
@@ -38,9 +38,18 @@ pub fn generate_for_peer(
     full_bleed: bool,
     host_display_name: Option<&str>,
 ) -> Html {
-    let peer_email = client.get_peer_email(key).unwrap_or_else(|| key.clone());
+    // Use display_name (updated via heartbeat) for the floating label.
+    // Fall back to session_id only if nothing is set yet.
+    let peer_display_name = client
+        .get_peer_display_name(key)
+        .filter(|n| !n.is_empty())
+        .unwrap_or_else(|| key.clone());
 
-    let is_host = host_display_name.map(|h| h == peer_email).unwrap_or(false);
+    // Keep email-based allow-list check using get_peer_email (unchanged behaviour).
+    let peer_email = client.get_peer_email(key).unwrap_or_else(|| key.clone());
+    let is_host = host_display_name
+        .map(|h| h == peer_display_name)
+        .unwrap_or(false);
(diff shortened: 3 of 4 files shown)


testradav:
Can you post a before and after screenshot please?

iamgabrielsoft:
There's an helm chat issue @testradav

testradav:
> There's an helm chat issue @testradav

I see that. Thanks.  Was trying to see the feature as I thought we already did display the username and the peer list.  That's why providing a screenshot with the PR is something useful thanks

testradav:
> > There's an helm chat issue @testradav
> 
> I see that. Thanks. Was trying to see the feature as I thought we already did display the username and the peer list. That's why providing a screenshot with the PR is something useful thanks

Never mind. Now I see this is on the tile.

testradav (commented):
PR #690 Review: "Show peer display name on video tile and peer list"
                                                                                                                                                                                                                                                      
  39 files changed, +991/-368 — This PR is doing too much. The stated goal is showing display names instead of session IDs, but it bundles unrelated migrations, models, proto changes, config changes, and formatting tweaks.                      
                                                   
  Critical / Blocking

  1. Protobuf wire-incompatible breaking change (generated code)
  The regenerated media_packet.rs has renumbered the MediaType enum — MEDIA_TYPE_UNKNOWN (value 0) was removed and all values shifted down (VIDEO is now 0 instead of 1, AUDIO is now 1 instead of 2, etc.). The VideoMetadata.codec field was also
  silently dropped. These changes don't appear in the .proto file diff, meaning the generated Rust code is out of sync with the proto definitions. Any deployed server/client running the old code will misinterpret packets from the new code. This
  alone should block the PR.

  2. Session ID storage key mismatch (bug)
  In yew-ui/src/components/attendants.rs:get_or_create_session_id():
  // Reads with literal key:
  if let Ok(Some(id)) = storage.get_item("session_id") { ... }
  // Writes with constant:
  let _ = storage.set_item(SESSION_ID_KEY, &new_id);  // SESSION_ID_KEY = "vc_session_id"
  These are different keys. The session ID will never be found on subsequent loads, generating a new one every time — defeating the purpose of a stable session ID.

  3. get_peer_email now returns session_id instead of email
  video_call_client.rs:559 — The function still called get_peer_email was changed to return peer.session_id instead of peer.email (which was renamed to display_name). This breaks the users_allowed_to_stream() allowlist check in
  canvas_generator.rs, which compares against email addresses.

  4. Host detection broken after name change
  canvas_generator.rs now compares host_display_name (set at join time from the email/username) against peer_display_name (mutable, updated via heartbeat). After a user renames themselves, these won't match and the host crown icon disappears.

  Should Not Be In This PR

  5. Unrelated database migrations — Three new migration files (20251221011540, 20251221011853, 20251221012015) adding meeting_protection_fields, meeting_owners table, and waiting_room_queue table. These are for a completely different feature.

  6. meeting_owner.rs model — A full new model with raw SQL queries for meeting ownership, unrelated to display names.

  7. DEFAULT_SALT constant — bcrypt salt added to actix-api/src/constants.rs, not used by this PR.

  8. Rename-bug-fix.md — Personal brainstorming/design doc committed to repo root. Should not be checked in.

  9. webTransportEnabled changed to false in yew-ui/scripts/config.js — Local dev config change that would affect all developers.

  10. Migration CASCADE added — 20260203000003 changed DROP CONSTRAINT to use CASCADE, which can silently drop dependent foreign keys and views. This is a separate concern that needs explicit review.

  Bugs / Code Quality

  11. Dioxus-ui gets no benefit — New fields are set to empty strings with no callback:
  session_id: String::new(),
  display_name: String::new(),
  on_peer_display_name_changed: None,

  12. Peer.email field left as comment — // pub email: String, should be fully removed, not commented out.

  13. Unused code added — PeerStatus::NameChanged variant is never constructed. is_meeting_metadata_packet in ws_chat_session.rs is #[allow(dead_code)]. session_id and display_name fields added to MediaPacket proto (fields 10, 11) but never
  populated on the send path.

  14. Debug-level logging at info level — host.rs has log::info!(">>> SaveChangeName triggered...") style debug breadcrumbs that should be removed or downgraded.

  15. Format-only changes scattered throughout — Many files have format!("{}", x) → format!("{x}") changes that inflate the diff and should be a separate PR.

  Recommendation

  Do not merge as-is. This needs to be split into smaller PRs:
  1. Proto changes + regeneration (done correctly, without renumbering existing enum values)
  2. Display name feature (client + yew-ui + dioxus-ui)
  3. Database migrations for meeting ownership (separate feature)
  4. Formatting cleanup (separate trivial PR)

  The session ID key bug and get_peer_email semantic change need to be fixed before the display name feature can work correctly.

dbmate/db/migrations/20251221011540_add_meeting_protection_fields.sql
```diff
@@ -0,0 +1,18 @@
+-- migrate:up
+
+ALTER TABLE meetings
+ADD COLUMN meeting_title VARCHAR(255) NULL,
```
darioalessandro, line 4: Why do we need DB changes to show peer display name on video tile and peer list @iamgabrielsoft ?

iamgabrielsoft: The PR came from the bug-fix-user-rename branch, that why its seems like I touched it. The pull was from the bug-rename branch. I'll put this PR on draft mode until the bug-rename is merged

PR #691, opened 2026-03-02
fix: add || true to kubectl jsonpath slot queries so cleanup and undeploy don't crash when a PR was never deployed

jboyd01:
## Problem

`kubectl get namespaces -o jsonpath='{.items[0]...}'` exits with code 1 when the items array is empty (i.e. no namespace matches the label selector — the PR was never deployed). The `2>/dev/null` suppressed the error message but not the exit code, so `bash -euo pipefail` killed the `Find deployment slot` step before the `[ -z "$SLOT" ]` check could run. This caused `pr-cleanup` to always fail on merged PRs that had no active preview, producing a spurious failure comment.

The same latent bug existed in `pr-undeploy`.

## Changes

**pr-cleanup.yaml**
- Add `|| true` to `SLOT=$(kubectl ...)` query — root cause fix
- Add `|| true` + empty-check guard to `PG_PASSWORD` retrieval; a missing secret previously caused a hard failure with no diagnostic output
- Add `outputs` block to `cleanup` job exposing `slot_existed` and `slot`
- Add `comment-success` job so the PR receives a comment when automatic cleanup actually removes a slot (previously only failures were reported)

**pr-undeploy.yaml**
- Same `|| true` fix to `SLOT=$(kubectl ...)` query
- Same `|| true` + empty-check guard to `PG_PASSWORD` retrieval

## Testing

- Close a PR that was never deployed → cleanup runs silently, no failure comment
- Close a PR with an active preview → cleanup removes namespace/DB, posts success comment to PR
- Comment `/undeploy` on a PR with no active deployment → posts "not found" comment, no failure


testradav (approved):
👍

PR #693, opened 2026-03-02
Update pull request template with new checklist items

testradav:
pull request template with new checklist items

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [x] Project / Infra

## Testing
- [ ] Tests written or updated
- [x] No tests needed

## Other
- [ ] Documentation updated
- [ ] Before/After Images provided

PR #689, opened 2026-02-27
dioxus app support

jboyd01:
support for dioxus UI running in Kubernetes.

* enable a list of domains for CORS
* build the dioxus & push image
* new helm chart for dioxus container

I staged this with the assumption that yew-ui is going away and boldly called this new chart "videocall-ui".  We could rename to something dioxus specific.   I've got this deployed internally and it needs testing and validation.

PR #695, opened 2026-03-02
fix: preserve meeting URL through OAuth redirect for Dioxus UI

jboyd01:
## Problem

When an unauthenticated user navigates directly to a meeting URL (e.g. `https://app-dx.vc2.fnxlabs.com/meeting/1`), they should be redirected back to that meeting after signing in. Instead, they were landing at the app root.

## Root Cause

**Dioxus 0.7's router calls `history.replaceState` on boot**, stripping query parameters not defined in the route. The previous approach stored the return URL as `?returnTo=<url>` in the `/login` query string. By the time the user clicked "Sign in", the Dioxus router had already stripped it — `window.location.search` was empty, so `do_login()` fell back to sending `returnTo={origin}/` (root) to the backend instead of `/meeting/1`.

This is a Dioxus 0.7 behaviour change, not a backend issue. The backend `validate_return_to()` was working correctly.

## Fix

**`dioxus-ui/src/pages/meeting.rs`** — on auth failure, write the current URL to `sessionStorage["vc_oauth_return_to"]` before navigating to `/login`. Includes explicit `log::warn!` if sessionStorage is unavailable so the failure is diagnosable rather than silent.

**`dioxus-ui/src/components/login.rs`** — `do_login()` reads and clears `vc_oauth_return_to` from sessionStorage first, falls back to URL search params (handles any direct `/login?returnTo=` navigations), then falls back to the app origin root.

`sessionStorage` is tab-scoped and survives the page reload triggered by `set_href`, but is not affected by Dioxus router URL normalisation. The key is cleared immediately after reading so there is no stale-state risk across sessions.

**`helm/global/us-east/meeting-api/values.yaml` + `helm/meeting-api/values.yaml`** — add `https://app-dx.videocall.rs` to `ALLOWED_REDIRECT_URLS`. The Dioxus UI deployment (PR #689) added the Dioxus frontend at a new origin but did not update the backend's redirect allowlist, so absolute `returnTo` URLs from that frontend were silently rejected and fell back to `AFTER_LOGIN_URL`.

## Testing

- Verified end-to-end on my k8s deployment: unauthenticated hit to `/meeting/1` → OAuth → lands back at `/meeting/1` ✓
- `meeting-api` unit tests: 81/81 pass (all OAuth/redirect tests included)
- `dioxus-ui`: `cargo check --target wasm32-unknown-unknown` clean

## Checklist

- [x] Bug is fixed and verified working
- [x] No new unit tests needed — backend redirect validation already has comprehensive test coverage; frontend sessionStorage logic is too simple and browser-only to warrant wasm-bindgen-test setup
- [x] Rebased onto current `main` (incorporates #688, #687, #693)


testradav (approved):
👍

PR #696, opened 2026-03-02
feature:  allow host to set waiting room option

testradav:
## big feature upgrade

- Allow host to set waiting room option for their meetings.
- Sets the stage for future meeting options to be set or unset

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [x] Breaking change: **requires DB schema migration, which automatically runs upon make**
- [ ] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [ ] Documentation updated
- [x] Before/After Images provided

### edit meeting option

image

### note: this option is replaced by #699 
image

### meeting options toggles at the start of the meeting

image

### DIFF
--- a/.github/workflows/pr-cleanup.yaml
+++ b/.github/workflows/pr-cleanup.yaml
@@ -1,7 +1,7 @@
 name: PR Preview Cleanup on Close
 
 on:
-  pull_request:
+  pull_request_target:
     types: [closed]
 
 permissions:
@@ -21,6 +21,9 @@ jobs:
   cleanup:
     name: Cleanup preview environment
     runs-on: ubuntu-latest
+    outputs:
+      slot_existed: ${{ steps.find-slot.outputs.exists }}
+      slot: ${{ steps.find-slot.outputs.slot }}
     steps:
       - name: Configure kubectl
         uses: digitalocean/action-doctl@v2
@@ -37,8 +40,10 @@ jobs:
           PR_NUM=${{ github.event.pull_request.number }}
 
           # Find which slot this PR is using
+          # || true: kubectl exits 1 when jsonpath hits an empty array (no match);
+          # suppress that so -e doesn't kill the script before the [ -z ] check.
           SLOT=$(kubectl get namespaces -l app=preview,pr=${PR_NUM} \
-            -o jsonpath='{.items[0].metadata.labels.slot}' 2>/dev/null)
+            -o jsonpath='{.items[0].metadata.labels.slot}' 2>/dev/null || true)
 
           if [ -z "$SLOT" ]; then
             echo "No preview environment found for PR #${PR_NUM}, nothing to cleanup"
@@ -71,15 +76,19 @@ jobs:
           echo "Dropping database: ${DB_NAME} (slot ${SLOT})"
 
           # Get postgres password from secret
-          PG_PASSWORD=$(kubectl get secret postgres-credentials -n preview-infra -o jsonpath='{.data.password}' | base64 -d)
+          # || true: prevents -e from exiting if the secret doesn't exist
+          PG_PASSWORD=$(kubectl get secret postgres-credentials -n preview-infra \
+            -o jsonpath='{.data.password}' 2>/dev/null | base64 -d || true)
 
-          # Drop database
-          kubectl exec -n preview-infra postgres-postgresql-0 -- \
-            env PGPASSWORD="${PG_PASSWORD}" psql -U postgres -c "DROP DATABASE IF EXISTS ${DB_NAME};" || {
-              echo "⚠️  Failed to drop database, may need manual cleanup"
-            }
-
-          echo "✅ Database dropped for slot ${SLOT}"
+          if [ -z "$PG_PASSWORD" ]; then
+            echo "⚠️  Failed to get postgres credentials, skipping database drop"
+          else
+            kubectl exec -n preview-infra postgres-postgresql-0 -- \
+              env PGPASSWORD="${PG_PASSWORD}" psql -U postgres -c "DROP DATABASE IF EXISTS ${DB_NAME};" || {
+                echo "⚠️  Failed to drop database, may need manual cleanup"
+              }
+            echo "✅ Database dropped for slot ${SLOT}"
+          fi
 
       - name: Summary
         if: steps.find-slot.outputs.exists == 'true'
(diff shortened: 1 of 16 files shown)


jboyd01:
## Code Review

### Security — mostly clean, one gap

**Authorization:** Correct. The PATCH handler checks `creator_id == email` before touching anything, using the same pattern as `delete_meeting`. Non-owner returns 403. No issues there.

**SQL injection:** None. All new queries use parameterized `$1/$2/$3` binds. The `format!()` calls only interpolate the compile-time `PARTICIPANT_COLUMNS` constant — not user input.

**Race condition:** The `FOR UPDATE` lock on the meetings row in `join_attendee` correctly serializes against `update_waiting_room_enabled`, which also updates that row. Two concurrent `join_attendee` calls don't block each other (they write to `meeting_participants`, not `meetings`), but that's correct — two simultaneous joiners both get the same waiting room decision atomically.

**Gap:** No test for unauthenticated PATCH (no session cookie → should 401). The `AuthUser` extractor handles it, but it's not verified.

---

### Tests — solid but with gaps

The 638-line test file covers the important behavioral paths well:
- Toggle on/off (owner), non-owner 403, not-found 404
- Auto-admit when WR off, waiting queue when WR on
- Bulk-admit on toggle-off, re-enable restores waiting behavior

**Missing:**
- Unauthenticated PATCH (no 401 test)
- Empty body `{}` PATCH — handler treats it as a no-op, returns current state, but this is untested and undocumented
- Host joining a meeting with `waiting_room_enabled: false` — should always get a token regardless (correct behavior, not tested)
- PATCH on a soft-deleted meeting (should 404)

---

### Code Quality — three real issues

**1. `upsert_attendee_admitted` is dead code.** Added to `participants.rs`, never called anywhere. `join_attendee` inlines equivalent SQL directly instead. Either use it or delete it.

**2. Optimistic UI toggle with no error rollback.** Both toggle button handlers (in `attendants.rs` and `meetings_list.rs`) flip the signal immediately then fire the API call:

```rust
waiting_room_toggle.set(new_val);  // flipped immediately
spawn_local(async move {
    if let Err(e) = update_meeting(...).await {
        log::error!("...");  // error eaten, UI left in wrong state
    }
});
```

If the API call fails, the toggle shows the wrong state and stays there. The user has no idea.

**3. Toggle widget is copy-pasted twice.** The iOS-style pill toggle appears in both `attendants.rs` and `meetings_list.rs` with near-identical code (only pixel sizes differ: 44/24px vs 40/22px). Should be a reusable Dioxus component.

---

### Architecture — one minor smell

The PATCH handler fetches the meeting row once for the ownership check, then `update_waiting_room_enabled` fetches it again inside the transaction. Two round-trips to the same row. The existing `soft_delete` avoids this by folding the ownership check into the SQL: `WHERE room_id = $1 AND creator_id = $2`. That cleaner pattern isn't used here.

---

### Dioxus vs yew-ui

Feature is correctly absent from yew-ui. The only yew change is adding `.meeting-edit-btn` CSS — 25 lines of styles for a button class that yew-ui doesn't actually render. Dead CSS in the yew bundle.

---

### Summary

| Issue | Severity |
|---|---|
| `upsert_attendee_admitted` is dead code, never called | Medium |
| Optimistic toggle has no error rollback — UI shows wrong state on API failure | Medium |
| Toggle widget copy-pasted twice; should be a reusable Dioxus component | Medium |
| PATCH fetches meeting row twice (auth check + inside transaction) | Low |
| No unauthenticated PATCH test (401 case) | Low |
| No test for empty-body PATCH `{}` (no-op behavior undocumented) | Low |
| `.meeting-edit-btn` CSS added to yew-ui but unused there | Low |
| Authorization correct, `FOR UPDATE` correctly serializes join vs toggle | ✅ |
| No SQL injection | ✅ |
| No `unwrap()` in production paths | ✅ |

testradav:
## Addressing Code Review Feedback                                                                                                                                                                        
                                                                                                      
  All issues from @jboyd01's [review](https://github.com/security-union/videocall-rs/pull/696#issuecomment-3991181840) have been addressed:                                                                 
                                                                                                                              
  ### Medium Severity                                                                                                                                                                                       
                                                                                                                                                                                                            
  - **`upsert_attendee_admitted` dead code** — Deleted from `meeting-api/src/db/participants.rs`. The `join_attendee()` function already handles both paths (waiting vs auto-admit) atomically with `FOR    
  UPDATE`.

  - **Optimistic toggle with no error rollback** — Both toggle handlers now revert the UI state on API failure:
    ```rust
    if let Err(e) = update_meeting(&meeting_id, new_val).await {
        log::error!("Failed to update waiting room: {e}");
        waiting_room_toggle.set(!new_val); // rollback
    }

  - Toggle widget copy-pasted twice — Extracted into a reusable ToggleSwitch component (dioxus-ui/src/components/toggle_switch.rs) with configurable width/height props. Both attendants.rs (44×24) and
  meetings_list.rs (40×22) now use it.

  Low Severity

  - PATCH fetches meeting row twice — Refactored update_meeting handler to attempt the UPDATE … WHERE creator_id = $2 first (folding in the ownership check). Only falls back to a SELECT on failure to
  distinguish 404 vs 403. Happy path is now a single round-trip.
  - No unauthenticated PATCH test — Added test_update_meeting_unauthenticated (expects 401).
  - No empty-body PATCH test — Added test_update_meeting_empty_body_noop (expects 200, returns current state unchanged).
  - No soft-deleted PATCH test — Added test_update_meeting_soft_deleted_returns_404 (expects 404).
  - .meeting-edit-btn CSS in yew-ui unused — Removed 25 lines of dead CSS from yew-ui/static/style.css. The class is only used by dioxus-ui, which has its own copy.

  Files Changed

  - meeting-api/src/db/participants.rs — Removed upsert_attendee_admitted dead code
  - meeting-api/src/routes/meetings.rs — Eliminated double-fetch in PATCH handler
  - meeting-api/tests/waiting_room_toggle_tests.rs — +3 tests (unauth, empty body, soft-deleted)
  - dioxus-ui/src/components/toggle_switch.rs — NEW reusable ToggleSwitch component
  - dioxus-ui/src/components/mod.rs — Register toggle_switch module
  - dioxus-ui/src/components/attendants.rs — Use ToggleSwitch, add error rollback
  - dioxus-ui/src/components/meetings_list.rs — Use ToggleSwitch, add error rollback
  - yew-ui/static/style.css — Removed unused .meeting-edit-btn CSS

  All three crates compile clean (meeting-api, videocall-ui-dioxus, videocall-ui).

dioxus-ui/src/components/attendants.rs
```diff
@@ -436,6 +438,39 @@ pub fn AttendantsComponent(
...
+                                    span { style: "font-size: 0.9rem;", "Waiting Room" }
+                                    crate::components::toggle_switch::ToggleSwitch {
+                                        enabled: waiting_room_toggle(),
+                                        on_toggle: {
+                                            let meeting_id = meeting_id_for_toggle.clone();
+                                            move |new_val: bool| {
+                                                waiting_room_toggle.set(new_val);
+                                                let meeting_id = meeting_id.clone();
+                                                wasm_bindgen_futures::spawn_local(async move {
+                                                    if let Err(e) = crate::meeting_api::update_meeting(&meeting_id, new_val).await {
```
darioalessandro, line 457: how about showing the error on the UI? right now If I am following the code properly the new value will be set on the UI regardless?

testradav: image

testradav: The error message in the list of meetings the host owns is also there but will be gone with #699 since it dedicates a page to the meeting options rather than displaying them on the list

PR #703, opened 2026-03-03
fix: repair Docker dev setup for fresh clones

jboyd01:
Our team has been having problems with running videocall locally in docker on Windows.  This is the result at my running through it trying to get it properly working in my environment.

## Summary

Four independent bugs prevented `make up` from working out of the box:

1. `Dockerfile.yew.dev` and `Dockerfile.dioxus.dev` were missing
   `git config --global --add safe.directory /app`, causing Nix to
   refuse the bind-mounted repo on WSL2 ("not owned by current user").
   `Dockerfile.actix.dev` already had this line; now all three match.

2. Makefile used bare `docker compose -f docker/docker-compose.yaml`
   which sets the project directory to `docker/`, so `.env` was looked
   up in `docker/` instead of the repo root. Fixed with `--env-file .env`.
   Added a `.env` target that auto-creates from the sample if missing.

3. `docker/.env-sample` was stale: missing `ENABLE_OAUTH`, `OAUTH_PROVIDER`,
   `OAUTH_ISSUER`, `OAUTH_REDIRECT_URL`, `AFTER_LOGIN_URL`,
   `ALLOWED_REDIRECT_URLS`, `JWT_SECRET`, and `ACTIX_UI_BACKEND_URL`.
   Without `OAUTH_ISSUER`, meeting-api panics on startup with no obvious
   error in `make up` output.

4. `yew-ui/scripts/config.js` had `webTransportEnabled: "true"` committed,
   which requires separate cert setup (`launch_chrome.sh`) and breaks the
   default WebSocket fallback for new developers.

Also updates the README Docker Setup section with `.env` setup instructions,
Google OAuth credential steps, the correct access URL, and platform notes
for Rancher Desktop users (Traefik on port 80).

## Test plan
- [ ] Fresh clone, `cp docker/.env-sample .env`, fill credentials, `make up`
- [ ] Verify yew-ui accessible at `http://localhost`
- [ ] Verify meeting-api does not crash on startup
- [ ] Verify `make up` without `.env` auto-creates it from sample with a message


testradav (approved):

yew-ui/scripts/config.js
```diff
@@ -4,7 +4,7 @@ window.__APP_CONFIG = Object.freeze({
   webTransportHost: "https://127.0.0.1:4433",
   oauthEnabled: "true",
   e2eeEnabled: "false",
-  webTransportEnabled: "true",
+  webTransportEnabled: "false",
```
testradav, line 7: Should this be true?  Is this going to affect all branches not just the Docker PR deployments, correct?

testradav (changes requested):

yew-ui/scripts/config.js
jboyd01: There are difficulties using web transport with insecure (self signed) certs.   See https://github.com/security-union/videocall-rs/wiki/WebTransport---TLS-Cert for details.  I did exploratory work (windows only) and abandoned  it.  TLDR, while webtransport can be used in Docker deployments, it either requires you to obtain valid SSL certs for webtransport, or you have to utilize specific BRowser flags to ignore/allow the insecure communication.

testradav: But should the default be true and if it's not possible to use web transport then gracefully fall back to web socket?  Otherwise mainline deployment are gonna need to set it to true every time main deployment gets updated

testradav (approved):
👍

yew-ui/scripts/config.js
testradav: I see these are just defaults and will be overwritten at one time if need be

PR #705, opened 2026-03-03
fix: performance improvements

testradav:
# Fix peer state propagation delays in multi-peer calls

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Enhancement
- [x] Breaking change -> rebuild your containters
- [ ] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [x] Documentation updated

## Summary

- **Send immediate heartbeat on mute/camera/screen toggle** — peers learn about state changes within one network round-trip instead of waiting up to 1 second for the next heartbeat tick
- **Increase diagnostics broadcast buffer (256 to 1024) and enable overflow mode** — prevents silent event drops under load; stale events are evicted instead of new ones being rejected
- **Consolidate per-peer diagnostics subscriptions into a single dispatcher** — replaces N independent async tasks (one per PeerTile) with one shared subscriber, eliminating O(N squared) event loop contention
- **Infer media-enabled state from actual frames** — receiving a VIDEO/AUDIO/SCREEN frame immediately sets the enabled flag, so the UI does not wait for a heartbeat to show video tiles
- **Only fire immediate heartbeats on actual state changes** — prevents redundant heartbeats from periodic re-affirmation (swap + compare instead of unconditional store)

## Context

During a ~10-person video call, three issues were observed:
1. Mute/unmute button on peer tiles not updating when peers toggled their mic
2. All videos appeared off despite peers having cameras on
3. A sudden burst of all videos appearing at once

Root causes identified:
- State changes only propagated via 1Hz heartbeat timer (up to 1s delay)
- 256-message diagnostics broadcast buffer overflowed silently with 10+ peers
- 10 independent PeerTile subscribers competing on the single-threaded browser event loop
- UI depended solely on heartbeats to learn about video/audio state, not actual media frames

## Test plan

- [ ] 2-person call: verify mute/unmute reflects immediately on peer tile (no visible delay)
- [ ] 2-person call: verify camera on/off reflects immediately on peer tile
- [ ] 5+ person call: verify all peer video tiles show correctly on join (no all off then burst on)
- [ ] 5+ person call: verify mute/unmute updates are responsive across all tiles
- [ ] Verify screen share start/stop reflects on peer tiles without delay
- [ ] Verify peer disconnect cleans up their entry (no stale tiles)
- [ ] Verify reconnection scenario: after connection drop + reconnect, state re-syncs correctly

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/actix-api/src/actors/session_logic.rs
+++ b/actix-api/src/actors/session_logic.rs
@@ -241,7 +241,8 @@ impl SessionLogic {
                 if self.observer {
                     trace!(
                         "Observer session {} dropping media packet from {}",
-                        self.id, self.email
+                        self.id,
+                        self.email
                     );
                     InboundAction::Processed
                 } else {


testradav:
Code review addressed

darioalessandro (approved):
Changes look reasonable, consider sharing test results

videocall-client/src/connection/connection.rs
```diff
@@ -194,6 +217,53 @@ impl Drop for Connection {
...
+    };
+
+    if let Some(sid) = session_id.borrow().as_ref() {
+        packet_wrapper.session_id = *sid;
+    }
+
+    Some(packet_wrapper)
+}
+
+fn aes_encrypt_heartbeat(aes: &Aes128State, packet: &MediaPacket) -> Result<Vec<u8>, ()> {
```
darioalessandro: Return the errors I think those log statements eat up the errors and return Error(void)

testradav:
Address code reviews

testradav:
this is promising.  
- substituted polling http for nats messaging where allowed for a substantial optimization of the client app
- all client errors addressed
- local testing working better than ever
- tests updated

image

darioalessandro:
This is elegant Tony!! Me likey!! Sent from my iPhoneOn Mar 5, 2026, at 15:03, Antonio Estrada ***@***.***> wrote:﻿testradav left a comment (security-union/videocall-rs#705)
this is promising.

substituted polling http for nats messaging where allowed
all client errors addressed
local testing working better than ever
tests updated

image.png (view on web)

testradav:
will look into screen share hic-ups.  good to go otherwise in our local test

PR #720, opened 2026-03-06
Fix/heartbeat display name

iamgabrielsoft:
This pull request describes and provide solution to bug fix for username renaming for issue https://github.com/security-union/videocall-rs/issues/452


testradav:
converting to draft, possibly closing unmerged.  replaced by #776

PR #724, opened 2026-03-08
fix: email, userid, session usage confusion

testradav:
- refactor session, email, username, display_name, and userid as per the plan in order to avoid confusion and misuses of variable names and values throughout which can lead to assumptions and conflicts
  - lots of files changed but mostly because of refactoring
- added peer "joins" and "leaves" NATs msg for immediate response (instead of waiting 5 to 10 secs for heartbeat cycle)
  - added entry/exit toast messages for attendants

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [x] Breaking change - data migration / schema change handled via make
- [ ] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [x] Documentation updated

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,7 +87,7 @@ Always delegate work to the specialized roster agents instead of making changes 
 - **frontend-rust-webtransport-and-websocket** — All Dioxus/Yew UI changes (components, pages, styling, state management)
 - **backend-rust-streaming** — All backend/API changes (Axum routes, DB queries, server logic)
 - **code-reviewer** — Review all code changes before committing
-- **web-security-auditor** — Review code touching auth, user input, API endpoints
+- **web-security-auditor** — Full-scope application security: backend auth/authz, API endpoints, input validation, XSS/injection, CSRF, UI trust indicators (e.g. host badges, role icons, permission displays), identity comparison logic, token handling, phishing vectors, architectural security review. Must audit both server-side AND client-side code — rendering code that conveys trust or authority is security-critical.
 - **database-reviewer** — Review schema, migration, and query changes
 - **integration-test-writer** — Write integration tests for new or changed features
 - **deploy-sync-expert** — Update Docker/K8s configs when services or dependencies change


testradav:
# Identity Cleanup: Rename `email` to `user_id` / `username` to `display_name`

**Branch:** `fix/uuid`
**76 files changed**, 1318 insertions, 870 deletions

---

## Summary

Renamed internal variables and protobuf fields to clearly separate three identity concepts:

| Concept | Before (variable names) | After |
|---------|------------------------|-------|
| **User identity** | `email`, `userid`, `user_id`, `sub`, `creator_id` | `user_id` everywhere |
| **Display name** | `username`, `email` (sometimes) | `display_name` |
| **Session** | `session_id` | `session_id` (unchanged) |

The `email` field in protobuf messages was misleading — it carried user identity (not always an email). Display name was used for identification in some places, which is fragile. This refactor enforces clean separation.

---

## Wire Compatibility

Protobuf field **numbers** are unchanged. Only field **names** changed. Binary wire format is fully backward-compatible. Old and new clients/servers can interoperate during rollout.

---

## Proto Schema Changes (5 files)

| File | Change |
|------|--------|
| `protobuf/types/packet_wrapper.proto` | `string email = 2` → `string user_id = 2` |
| `protobuf/types/media_packet.proto` | `string email = 2` → `string user_id = 2`; added `bool is_speaking = 4` to `HeartbeatMetadata` |
| `protobuf/types/meeting_packet.proto` | `string target_email = 7` → `string target_user_id = 7` |
| `protobuf/types/health_packet.proto` | `string reporting_peer = 3` → `string reporting_user_id = 3` |
| `protobuf/types/server_connection_packet.proto` | `string customer_email = 2` → `string user_id = 2` (in `ConnectionMetadata`) |

Generated Rust types were regenerated from protos via `cd protobuf && make build`.

---

## Generated Types (5 files)

| File | Notes |
|------|-------|
| `videocall-types/src/protos/packet_wrapper.rs` | Regenerated — `.email` → `.user_id` |
| `videocall-types/src/protos/media_packet.rs` | Regenerated — `.email` → `.user_id`, added `is_speaking` field |
| `videocall-types/src/protos/meeting_packet.rs` | Regenerated — `.target_email` → `.target_user_id` |
| `videocall-types/src/protos/health_packet.rs` | Regenerated — `.reporting_peer` → `.reporting_user_id` |
| `videocall-types/src/protos/server_connection_packet.rs` | Regenerated — `.customer_email` → `.user_id` |

---

## Shared Type Crates (4 files)

| File | Change |
|------|--------|
| `videocall-types/src/lib.rs` | `SYSTEM_USER_EMAIL` → `SYSTEM_USER_ID` |
| `videocall-meeting-types/src/responses.rs` | `ParticipantStatusResponse.email` → `.user_id` |
| `videocall-meeting-types/src/requests.rs` | `AdmitRequest.email` → `.user_id` |
| `videocall-meeting-types/src/token.rs` | Updated doc comment on `sub` claim |

---

## Media Server — actix-api (13 files)

| File | Key Changes |
|------|-------------|
| `actors/session_logic.rs` | Type alias `Email` → `UserId`; field `email` → `user_id` |
| `actors/mod.rs` | Re-export `UserId` |
| `actors/chat_server.rs` | `SYSTEM_USER_EMAIL` → `SYSTEM_USER_ID`; message struct fields; test assertions |
| `actors/transports/ws_chat_session.rs` | `email` param → `user_id` |
| `actors/transports/wt_chat_session.rs` | `email` param → `user_id` |
| `session_manager.rs` | `SYSTEM_USER_EMAIL` → `SYSTEM_USER_ID`; `ReservedUserEmail` → `ReservedUserId` |
| `lobby.rs` | JWT extraction: `email` → `user_id` throughout |
| `token_validator.rs` | Doc comments updated |
| `server_diagnostics.rs` | `customer_email` → `user_id` |
| `client_diagnostics.rs` | `reporting_peer` → `reporting_user_id` |
| `bin/metrics_server.rs` | `reporting_peer` → `reporting_user_id` |
| `bin/metrics_server_snapshot.rs` | `customer_email` → `user_id` |
| `webtransport/mod.rs` | Log messages and test assertions |

---

## Meeting API (11 files)

| File | Key Changes |
|------|-------------|
| `auth.rs` | `AuthUser.email` → `.user_id` |
| `token.rs` | Function param `email` → `user_id` |
| `error.rs` | Param name updated |
| `nats_events.rs` | `SYSTEM_EMAIL` → `SYSTEM_USER_ID`; `target_email` → `target_user_id` |
| `db/participants.rs` | Struct field, SQL column refs, function params |
| `routes/participants.rs` | `AuthUser { email }` → `{ user_id }` throughout |
| `routes/meetings.rs` | Same pattern |
| `routes/waiting_room.rs` | Same pattern |
| `routes/oauth.rs` | `get_profile` destructure updated |
| `tests/observer_token_tests.rs` | JSON `"email"` → `"user_id"`; assertions `.email` → `.user_id` |
| `tests/participant_tests.rs` | Same |
| `tests/waiting_room_tests.rs` | Same |

---

## Client Library — videocall-client (10 files)

| File | Key Changes |
|------|-------------|
| `client/video_call_client.rs` | `VideoCallClientOptions.userid` → `.user_id`; `get_peer_email()` → `get_peer_user_id()`; `SYSTEM_USER_EMAIL` → `SYSTEM_USER_ID`; all `.email` on protos → `.user_id` |
| `connection/connection.rs` | Heartbeat packet: `email:` → `user_id:` |
| `connection/connection_manager.rs` | Self-filter: `packet.email` → `packet.user_id` |
| `encode/transform.rs` | `MediaPacket` and `PacketWrapper` field renames |
| `encode/microphone_encoder.rs` | Same |
| `encode/camera_encoder.rs` | Same |
| `encode/screen_encoder.rs` | Same |
| `decode/peer_decode_manager.rs` | `Peer.email` → `.user_id`; method params |
| `health_reporter.rs` | `reporting_peer` → `reporting_user_id` |
| `lib.rs` | Doc comment updates |

---

## Dioxus UI (15 files)

| File | Key Changes |
|------|-------------|
| `src/context.rs` | `UsernameCtx` → `DisplayNameCtx`; storage key `vc_username` → `vc_display_name`; functions `load/save/clear_username_*` → `*_display_name_*`; added `MeetingHost` struct |
| `src/main.rs` | Updated imports and context provider |
| `src/pages/home.rs` | Variable renames `username_*` → `display_name_*`; storage function calls |
| `src/pages/meeting.rs` | `current_user_email` → `current_user_id`; response `.email` → `.user_id` |
| `src/components/attendants.rs` | `email` prop → `user_id`; `VideoCallClientOptions { userid: }` → `{ user_id: }` |
| `src/components/host_controls.rs` | `participant.email` → `.user_id` |
| `src/components/peer_list.rs` | `get_peer_email` → `get_peer_user_id`; context type |
| `src/components/canvas_generator.rs` | `get_peer_email` → `get_peer_user_id` |
| `src/components/diagnostics.rs` | Same |
| `src/components/host.rs` | Storage function renames |
| `src/components/waiting_room.rs` | `email` prop → `user_id` |
| `README.md` | `UsernameCtx` → `DisplayNameCtx` |
| `tests/home_integration.rs` | Imports + storage key updated |
| `tests/context_unit.rs` | Function names + storage key updated |

---

## Yew UI (7 files — minimal compatibility updates)

| File | Key Changes |
|------|-------------|
| `components/canvas_generator.rs` | `get_peer_email` → `get_peer_user_id` |
| `components/peer_list.rs` | Same |
| `components/host_controls.rs` | `p.email` → `p.user_id` on response types |
| `components/attendants.rs` | `userid:` → `user_id:` in options |
| `components/waiting_room.rs` | Same |
| `pages/meeting.rs` | Response `.email` → `.user_id`; options `userid:` → `user_id:` |
| `pages/home.rs` | `profile.email` → `profile.user_id` |

---

## CLI (3 files)

| File | Key Changes |
|------|-------------|
| `videocall-cli/src/producers/camera.rs` | Proto field renames |
| `videocall-cli/src/producers/microphone.rs` | Same |
| `videocall-cli/src/consumers/webtransport.rs` | Same |

---

## Database Migration (1 new file)

**`dbmate/db/migrations/20260307000001_rename_email_to_user_id.sql`**

```sql
-- migrate:up
ALTER TABLE meeting_participants RENAME COLUMN email TO user_id;

-- migrate:down
ALTER TABLE meeting_participants RENAME COLUMN user_id TO email;
```

---

## OAuth Boundary — NOT Renamed

These use the OIDC standard `email` claim from identity providers and were correctly left as-is:

- `meeting-api/src/oauth.rs` — `IdTokenClaims.email`
- `meeting-api/src/routes/oauth.rs` — `claims.email` extraction from provider
- `actix-api/src/auth/mod.rs` — OAuth claims processing
- `actix-api/src/bin/websocket_server.rs` — OAuth cookie handling

---

## Remaining Work (deferred)

1. **`host_user_id` in API responses** — `MeetingHost` context exists in dioxus-ui but `host_user_id` is not yet in `ParticipantStatusResponse` or populated by meeting-api routes
2. **Non-OAuth UUID generation** — `get_or_create_local_user_id()` not yet implemented; users without OAuth still lack stable identity
3. **localStorage migration** — Old `vc_username` key not auto-migrated to `vc_display_name` for returning users

---

## Compilation Verified

| Crate | Target | Status |
|-------|--------|--------|
| `videocall-types` | native | Pass |
| `videocall-meeting-types` | native | Pass |
| `meeting-api` (+ all tests) | native | Pass |
| `videocall-api` (actix-api) | native | Pass |
| `videocall-client` (default) | wasm32 | Pass |
| `videocall-client` (no-default-features) | wasm32 | Pass |
| `videocall-cli` | native | Pass |
| `videocall-ui-dioxus` | wasm32 | Pass |
| `videocall-ui` (yew) | wasm32 | Pass |


---

## Fix Race condition when joining meeting: waiting for admission vs early admission

The Race                                                                                                                                                                                                
                                                                                                                                                                                                        
  T0: User clicks Join → join_meeting() HTTP call                                                                                                                                                         
  T1: API returns status="waiting" + observer_token                                                                                                                                                       
      → Host is notified immediately via NATS                                                                                                                                                             
  T2: Client renders WaitingRoom component                                                                                                                                                                
  T3: WaitingRoom use_effect fires                                                                                                                                                                        
  T4: VideoCallClient::new() — no connection yet                                                                                                                                                          
  T5: client.connect() — starts WebSocket/WebTransport handshake                                                                                                                                          
                                                                                                                                                                                                          
      ⚡ HOST ADMITS USER HERE (NATS event published)                                                                                                                                                     

  T6: Observer WebSocket actually connects and subscribes
      → PARTICIPANT_ADMITTED event already gone — NATS doesn't replay
      → Client stuck in waiting room                                                                                                                                                                      
                                                                                                                                                                                                          
  The client does have a polling fallback every 5 seconds (waiting_room.rs:159-234), but the polling only runs when the observer is NOT connected. Once the observer connects (T6), polling stops — but the admission event was already missed.

testradav:
fixing 4 additional bugs:

# Bug 1: encoderWorker.min.js crash on mic off                    

  File: videocall-client/src/audio_worklet_codec.rs:158-163                                                                                                                                               

  The destroy() method sends messages in this order: Close → Flush → Done. But the worker code shows:                                                                                                     
  - Flush calls this.encoder.flush() — can fail if encoder is already stopped                                                                                                                             
  - Done calls this.encoder.encodeFinalFrame().forEach(...) — if encodeFinalFrame() returns undefined, .forEach() crashes

  The correct order should be Flush → Done → Close (flush pending data, finalize, then close). And Done already calls this.encoder.destroy() and delete this.encoder, so sending Flush after Done would   
  also fail.                                                                                                                                                                                              

  Fix: Reorder to Flush → Done → Close, or better yet just send Done (which internally flushes and destroys).                                                                                             

# Bug 2: peer_tile.rs panic — Signal dropped before async task                                                                                                                                            

  File: dioxus-ui/src/components/peer_tile.rs:37-81, 123                                                                                                                                                  

  The panic: called Result::unwrap() on an Err value: Dropped(ValueDroppedError) at line 123 (which is inside handle_diagnostics_event calling .peek() on a signal).                                      

  The problem: use_effect spawns an async task via wasm_bindgen_futures::spawn_local (line 78) that holds Signal handles (audio_enabled, video_enabled, etc.). When the PeerTile component unmounts (e.g.,
   peer leaves), the signals created by use_signal at lines 37-40 are dropped. But the spawned async task continues running and tries to .peek() the dropped signals at lines 123/128/133/138, causing the
   panic.                                                                                                                                                                                                 

  The Abortable mechanism (lines 47-50) only aborts on re-render, not on unmount. The abort handle stored in use_hook is never cleaned up on component drop.                                              

  Fix: Use use_drop or use_on_destroy to abort the subscription when the component unmounts. Or use Dioxus spawn instead of spawn_local (which ties the future's lifetime to the component scope).        

# Bug 3: (Potential) spawn_local instead of spawn in peer_tile                                                                                                                                            

  This is related to Bug 2 — line 78 uses wasm_bindgen_futures::spawn_local instead of Dioxus's spawn. This means the async task outlives the component, which is the root cause of the dropped signal    
  panic. The same pattern was already fixed in home.rs for the navigator issue.                                                                                                                           

# 4. Yew-UI missing features from identity cleanup

  - No get_or_create_local_user_id() function
  - No migrate_legacy_storage() function
  - No USER_ID_STORAGE_KEY constant
  - normalize_spaces re-exported from yew-ui but removed from dioxus-ui exports

testradav:
```
  ┌─────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────┐                                                                                 
  │                    File                     │                                  Fix                                  │                                                                                 
  ├─────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────┤                                                                                 
  │ dioxus-ui/src/components/peer_tile.rs       │ spawn_local → spawn — prevents Signal panic on unmount                │
  ├─────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────┤
  │ videocall-client/src/audio_worklet_codec.rs │ destroy() reordered: Flush → Done → Close — prevents JS forEach crash │
  └─────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────┘

  Missed Renames (from code review)

  ┌───────────────────────────────────────────────────────┬──────────────────────────────────────────┐
  │                         File                          │                   Fix                    │
  ├───────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ protobuf/types/rsa_packet.proto                       │ username → user_id                       │
  ├───────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ videocall-types/src/protos/rsa_packet.rs              │ Generated file updated                   │
  ├───────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ videocall-client/src/client/video_call_client.rs:1048 │ username: → user_id: in RsaPacket        │
  ├───────────────────────────────────────────────────────┼──────────────────────────────────────────┤
  │ videocall-meeting-client/src/lib.rs:34                │ Doc-test profile.email → profile.user_id │
  └───────────────────────────────────────────────────────┴──────────────────────────────────────────┘

  Yew-UI Consistency (12+ files)

  Full rename: UsernameCtx → DisplayNameCtx, vc_username → vc_display_name, host_email → host_user_id, storage functions renamed, user_email → user_id, email prop → display_name/user_id, added UUID
  generation + legacy migration.

  Both UIs — email prop cleanup

  ┌───────────────────────────────────────────┬──────────────────────────────────────────────────────────────────┐
  │                   File                    │                               Fix                                │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ dioxus-ui/src/components/attendants.rs    │ user_email → user_id, duplicate user_id → display_name + user_id │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ dioxus-ui/src/components/meetings_list.rs │ user_email → user_id                                             │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ dioxus-ui/src/pages/meeting.rs            │ Updated prop call sites                                          │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ yew-ui/src/components/attendants.rs       │ Same pattern — email → display_name + user_id                    │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ yew-ui/src/components/waiting_room.rs     │ email → user_id                                                  │
  ├───────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┤
  │ yew-ui/src/components/meetings_list.rs    │ user_email → user_id                                             │
  └───────────────────────────────────────────┴──────────────────────────────────────────────────────────────────┘

```

meeting-api/src/nats_events.rs
```diff
@@ -22,14 +22,14 @@ use videocall_types::protos::meeting_packet::MeetingPacket;
 use videocall_types::protos::packet_wrapper::packet_wrapper::PacketType;
 use videocall_types::protos::packet_wrapper::PacketWrapper;
 
-/// System email used as the sender for server-generated NATS packets.
-const SYSTEM_EMAIL: &str = "system-&^%$#@!";
+/// System user ID used as the sender for server-generated NATS packets.
+const SYSTEM_USER_ID: &str = "system-&^%$#@!";
```
darioalessandro: @testradav use the constant in videocall-types https://github.com/security-union/videocall-rs/blob/176ddafbe22d1ad94580935ce3ba1718f4a6ca69/videocall-types/src/lib.rs#L36

darioalessandro (changes requested):
Hey @testradav I love that you are tackling this at the root instead of carrying forward the tech deb! I rather breaking it once than living forever with the issues as we scale. 

Please pack the user id as bytes so that we use 16 bytes a 128 bit UUID as opposed to 32 bytes if we use a string. I would add the change myself but because it is a fork from your repo I do not have permissions.

Claude should be able to plow through this change in no time!!!

meeting-api/src/nats_events.rs
testradav: addressing... 

...the userid displayed in parenthesis on entry/exit messages, and the names on the tiles of the participants are displaying some sort of uuid string instead of the expected value.  example, with this identity service, the user id is actually an email address.

⏺ That's the core issue with the UUID v5 approach — it's a one-way hash, so the original email is irrecoverably lost. When the UI reads user_id_bytes_to_string(), it gets a UUID hex string instead of the email. 

Since user_ids are currently emails (not UUIDs), the right fix is to store them as raw UTF-8 bytes in the proto bytes fields — not UUID v5 hashed. This:                                                
  - Keeps the proto field type as bytes (satisfying the reviewer's structural change)
  - Preserves the original email for display  
  - Can be upgraded to actual 16-byte UUIDs later when a proper identity system is built
                                                                                                                                                                                                  
The fix: replace all `to_user_id_bytes(&email) → email.as_bytes().to_vec() and all user_id_bytes_to_string(&bytes) → String::from_utf8_lossy(&bytes).to_string()`

jboyd01:
Hey @testradav 

Really fantastic to be addressing this.  Claude review calls out 3 issues he'd like to see fixed and a couple of potential issues that remain:

 Critical Issues (Must Fix Before Merge)

  1. **creator_id repurposed for display_name in MeetingPacket**
  `build_peer_joined_packet / build_peer_left_packet` in session_manager.rs stuff display_name into creator_id. That field means "meeting creator" for MEETING_STARTED events. Any consumer processing both event types must now know creator_id means different things depending on event_type. A dedicated proto field is the right fix.

  2. **is_host comparison: display_name compared against user_id**
  In canvas_generator.rs (both Dioxus and Yew):
  `let is_host = host_display_name.map(|h| h == peer_user_id).unwrap_or(false);`
  This compares a display name against a user_id — two concepts this PR explicitly separates. Will always return false once display names diverge from user IDs. Should be peer_display_name or introduce host_user_id.

  3. **can_stream checks display_name instead of user_id**
  `let can_stream = is_allowed.iter().any(|host| host == &display_name);`
  The allowed-to-stream list is keyed by identity, but the comparison is against the human-readable display name. Users will be incorrectly denied streaming whenever their display name
  doesn't match their user_id.

  _regarding 3, I'm not certain this is setting is used anymore.  @darioalessandro do you know about this one off hand? **can_stream** ?_

  ---
  Moderate Issues

  Double join sound — Msg::OnPeerAdded and Msg::OnPeerJoined both call play_user_joined() in the Yew UI. Every peer join will play the chime twice. OnPeerAdded sound should be removed now
  that OnPeerJoined covers it.

  Toast parsing by string-contains — msg.contains("joined") / msg.contains("left") to determine toast type. A display name containing "left" breaks this. Should be a typed enum.

  AudioContext leak — play_tone_pair creates a new AudioContext on every join/leave without closing it. Browsers limit concurrent AudioContexts (typically 6). High-churn meetings will
  exhaust it.

testradav:
Reviewer Comments Status                                                                                                                                                                             
                                                         
  1. SYSTEM_USER_ID constant — ADDRESSED. Now imports from videocall_types instead of local constant.
  2. UUID binary packing — NOT FULLY ADDRESSED. Current implementation uses as_bytes() on UTF-8 strings (same size). True 16-byte UUID packing not implemented. Commit says "uuid packing" but it's just 
  string→bytes wire type change.                                                                                                                                                                         
                                                                                                                                                                                                         
  user_id vs display_name Audit — CLEAN                                                                                                                                                                  
                                                                                                                                                                                                       
  All 11 areas audited show correct separation. One minor design note: can_stream in attendants.rs falls back to display_name when user_id is None, but in practice user_id is always provided.          
                                                                                                                                                                                                       
  Performance — NO REGRESSIONS                                                                                                                                                                           
                                                                                                                                                                                                       
  - Hot path (per-frame media): unchanged                                                                                                                                                                
  - Memory: +~5KB total (room_members + display_name_cache), properly cleaned up
  - New NATS events: join/leave only, ~200 bytes each, negligible

jboyd01:
Tony, one concern: The host crown icon was determined by comparing host_display_name against each peer's display_name. Display names are user-chosen free text — they are not unique identifiers. Any participant with the same display name as the host would incorrectly receive the crown on every other participant's screen.

See https://github.com/testradav/videocall-rs/pull/173  which merges a fix into this branch.  Everything else looks solid, no concerns outside of this.

testradav:
```
  ┌──────────┬─────────────────────────────────────────────────────────┐                                                                                                                                 
  │  Commit  │                       Description                       │
  ├──────────┼─────────────────────────────────────────────────────────┤                                                                                                                                 
  │ bfac2fac │ Fix relay test: skip PARTICIPANT_JOINED packets         │                              
  ├──────────┼─────────────────────────────────────────────────────────┤
  │ 3f9aa323 │ Add polling fallbacks for waiting room reliability      │                                                                                                                                 
  ├──────────┼─────────────────────────────────────────────────────────┤
  │ 5d340920 │ Security: use user_id (not display_name) for host crown │                                                                                                                                 
  ├──────────┼─────────────────────────────────────────────────────────┤                                                                                                                                 
  │ fd6c1388 │ Expand web-security-auditor scope in CLAUDE.md          │
  └──────────┴─────────────────────────────────────────────────────────┘                                                                                                                                 
```

protobuf/types/health_packet.proto
```diff
@@ -50,7 +50,7 @@ message PeerStats {
 message HealthPacket {
   string session_id = 1;
   string meeting_id = 2;
-  string reporting_peer = 3;
+  bytes reporting_user_id = 3;
```
darioalessandro, line 53: nice!!

protobuf/types/meeting_packet.proto
```diff
@@ -26,11 +26,14 @@ message MeetingPacket {
...
-  string creator_id = 5;
+  bytes creator_id = 5;
   int64 participant_count = 6;
-  // For ADMITTED/REJECTED: affected user's email
-  string target_email = 7;
+  // For ADMITTED/REJECTED: affected user's ID
+  bytes target_user_id = 7;
   // For ADMITTED: JWT to upgrade connection
   string room_token = 8;
+  uint64 session_id = 9;
```
darioalessandro, line 35: This session id is dupped, we have another session_id in the packet_wrapper that we can use

PR #755, opened 2026-03-12
performance optimization

testradav:
# Big Performance Optimization

Plan/Thesis: https://github.com/testradav/videocall-rs/blob/perf-opt/docs/PERFORMANCE_OPTIMIZATION_PLAN.md

### Objective: Make videocall adapt to network conditions on-the-fly in order to support wider array of deployments and communication situations gracefully

- P0: Add adaptive quality constants file and migrate hardcoded values from encoders, PID controller, and connection to centralized config
- P1: Add adaptive quality state machine with automatic video/audio tier selection and integrate PLI keyframe requests, framerate/resolution adaptation, and audio bitrate adaptation
- P2: Add automatic reconnection with exponential backoff, connection quality re-election, event-driven heartbeats, and VAD mute optimization
- P3: Add visibility-based decode optimization and WebTransport datagram support for media packets
- P4: Add Opus in-band FEC, audio/video lip-sync, VP9 SVC temporal layers, server congestion feedback, and redundant audio packets

## Security Fixes

- S1: Server drops client-originated CONGESTION packets — new PacketKind::Dropped + is_congestion_packet() check
- S2: Server rate-limits KEYFRAME_REQUEST — KeyframeRequestLimiter allows max 2/sec per session, excess dropped with warning
- S3: CONGESTION routing changed from broadcast (room.{room}.congestion) to targeted (room.{room}.{sender_sid}) — fixes information leakage + misleading comment
- S4: Reconnection fast-fails after 3 consecutive zero-connection attempts — stops wasting retries against revoked sessions
- 7 new server-side unit tests for packet classification and rate limiting

## Performance Fixes

Performance:
- P2: Eliminated duplicate AdaptiveQualityManager — camera encoder shares audio tier atomics with microphone encoder, removed redundant diagnostics processing
- P3: Default video tier changed from "high" (720p) to "medium" (480p) — avoids wasted encoding on constrained devices during initial step-down window

## Code quality:

- Q1: Fixed Dioxus IntersectionObserver memory leak — observers now stored in Signal, disconnected on unmount via use_drop
- Q2: Changed fallback try_borrow() to borrow() — surfaces logic errors instead of silently creating orphan flags
- Q3: Added stale sender cleanup to CongestionTracker — entries older than 10× congestion window removed opportunistically
- Q4: Confirmed screen share uses video_metadata.sequence — added clarifying comment
- Q5: Replaced hardcoded 20ms RED frame duration with OPUS_FRAME_DURATION_MS constant
[...]

PR #759, opened 2026-03-14
perf 02: server-side hardening - protobuf types, congestion tracking, datagram routing

testradav:
## Summary
- Adds `KEYFRAME_REQUEST` and `CONGESTION` protobuf message types for client-server signaling
- Server drops client-originated CONGESTION packets (prevents forgery)
- Rate-limits KEYFRAME_REQUEST to max 2/sec per session (prevents flooding)
- Changes congestion routing from broadcast to targeted (room.{room}.{sender_sid})
- Adds `CongestionTracker` with windowed drop counting, rate-limited notifications, stale entry cleanup
- Enables datagram-based send routing for WebTransport in `wt_chat_session`
- Adds `PacketKind::Dropped` classification for security-filtered packets

## Tests
- 24 unit tests: 11 packet handler (classification, rate limiting), 13 congestion tracker

## PR Stack (replaces #755)
1. Foundation: constants + quality manager (#758, merged first)
2. **This PR** - Server hardening
3. Connection resilience: reconnection + re-election + heartbeats
4. Decoder: PLI keyframe requests + RED audio + visibility
5. Encoder adaptation + client wiring + UI integration

### DIFF
--- a/actix-api/src/actors/chat_server.rs
+++ b/actix-api/src/actors/chat_server.rs
@@ -33,6 +33,7 @@ use protobuf::Message as ProtobufMessage;
 use std::collections::HashMap;
 use tokio::task::JoinHandle;
 use tracing::{error, info, trace, warn};
+use videocall_types::protos::packet_wrapper::packet_wrapper::PacketType;
 use videocall_types::protos::packet_wrapper::PacketWrapper;
 use videocall_types::SYSTEM_USER_ID;
 
@@ -541,7 +542,15 @@ fn handle_msg(
 ) -> impl Fn(async_nats::Message) -> Result<(), std::io::Error> {
     move |msg| {
         if msg.subject == format!("room.{room}.{session}").replace(' ', "_").into() {
-            return Ok(());
+            // Self-skip prevents echo of our own broadcasts. However,
+            // CONGESTION signals published on our subject by a congested
+            // receiver must still be delivered — they are not echo.
+            let is_congestion = PacketWrapper::parse_from_bytes(&msg.payload)
+                .map(|pw| pw.packet_type == PacketType::CONGESTION.into())
+                .unwrap_or(false);
+            if !is_congestion {
+                return Ok(());
+            }
         }
         let message = Message {
             msg: msg.payload.to_vec(),


jboyd01 (changes requested):
Looks solid.  Claude has but one complaint:

`classify_packet()` now calls `is_congestion_packet()` first, which does `PacketWrapper::parse_from_bytes(data)` on every single inbound packet. If the packet is not CONGESTION, it then proceeds to `is_rtt_packet()` (another parse) and potentially `is_health_packet()` (another parse). In the hot path (media packets at 30+ FPS per participant), this means **3 full protobuf deserializations per inbound packet** before it reaches the `Data` classification. This is a meaningful CPU cost on the relay server under load. Consider either:
  - Parsing once and passing the parsed `PacketWrapper` to classification functions, or
  - Using a fast byte-prefix check (e.g., checking the protobuf field tag for `packet_type` directly) before full parsing.
- **`is_keyframe_request()` double-parses**: After `classify_packet()` returns `Data`, `handle_inbound()` calls `is_keyframe_request(data)` which parses the `PacketWrapper` *again* plus the inner `MediaPacket`. For media packets (the vast majority), this is wasted work. This means a normal media packet triggers 4 protobuf parses total.

testradav:
> Looks solid. Claude has but one complaint:
> 
> `classify_packet()` now calls `is_congestion_packet()` first, which does `PacketWrapper::parse_from_bytes(data)` on every single inbound packet. If the packet is not CONGESTION, it then proceeds to `is_rtt_packet()` (another parse) and potentially `is_health_packet()` (another parse). In the hot path (media packets at 30+ FPS per participant), this means **3 full protobuf deserializations per inbound packet** before it reaches the `Data` classification. This is a meaningful CPU cost on the relay server under load. Consider either:
> 
> * Parsing once and passing the parsed `PacketWrapper` to classification functions, or
> * Using a fast byte-prefix check (e.g., checking the protobuf field tag for `packet_type` directly) before full parsing.
> * **`is_keyframe_request()` double-parses**: After `classify_packet()` returns `Data`, `handle_inbound()` calls `is_keyframe_request(data)` which parses the `PacketWrapper` _again_ plus the inner `MediaPacket`. For media packets (the vast majority), this is wasted work. This means a normal media packet triggers 4 protobuf parses total.

addressed here https://github.com/security-union/videocall-rs/pull/759/commits/e9a0b62a120a0a147a3e47d5ec9e6abcddbd1a2b

testradav:
> Critical (1)
> 
> 1. Wrong session ID passed to congestion tracker — In wt_chat_session.rs, msg.session is the receiver's own session ID, not the sender's. So on_outbound_drop() records drops against the receiver, and the CONGESTION notification published to NATS gets addressed to the receiver — the actual sender never sees it. The sender's session ID lives inside the serialized PacketWrapper.session_id in the payload. Fix: extract sender ID from the payload, or modify the Message struct to carry it separately.
> 
> Important (5)
> 
> 2. Outbound hot-path protobuf re-parse — should_use_datagram() calls PacketWrapper::parse_from_bytes(data) on every outbound packet to every recipient, just to check packet_type. In a 10-person call, that's a lot of redundant parsing for a "perf" PR. Could check the wire format field tag directly or pass packet type as metadata.
> 3. WebSocket sessions have no congestion tracking — WS uses ctx.binary() with no drop detection. CongestionTracker and KeyframeRequestLimiter fields in SessionLogic are dead weight for WS clients. May be intentional (TCP backpressure), but should be documented.
> 4. is_congestion_packet, is_keyframe_request, is_rtt_packet are dead code outside tests - they duplicate classify_packet() logic and do redundant full parses. Should be #[cfg(test)] or documented for future use.
> 5. DATAGRAM_MAX_SIZE in packet_handler.rs - Transport-layer constant belongs in constants.rs with the others.
> 6. Stale entry cleanup runs on every drop event - CongestionTracker::record_drop() does retain() over all entries on every call. O(n) per drop under congestion. Should be periodic instead.
> 
> Minor (2)
> 
> 7. classify_packet returns PacketKind::Data for unparseable packets — forwards garbage to all room participants. A PacketKind::Invalid that gets dropped would be safer.
> 8. on_outbound_drop spawns unbounded tokio::spawn per congestion notification — could accumulate tasks under sustained congestion.

addressed in https://github.com/security-union/videocall-rs/pull/759/commits/7373940575bd706853e1f037fc8dfb195f5f44c5

```
- Session ID bug (critical): 
  - Now extracts the sender's session_id from the PacketWrapper payload instead of using msg.session (which is the receiver's ID). 
  - CONGESTION signals now reach the correct sender.
- Outbound hot-path parse eliminated:
  - send_auto now takes an is_media: bool hint computed from the already-parsed PacketWrapper, instead of calling should_use_datagram() which re-parsed every packet.
- Dead code cleanup: 
  - is_congestion_packet(), is_rtt_packet(), is_keyframe_request(), should_use_datagram() moved to #[cfg(test)] — they duplicated classify_packet() logic and were only used by tests.
- Amortized cleanup: 
  - CongestionTracker::record_drop() now runs retain() every 100 drops instead of on every single drop.
```

testradav:
@darioalessandro ready for review

actix-api/src/actors/packet_handler.rs
```diff
@@ -42,82 +41,60 @@ pub enum PacketKind {
...
+/// parsed at most once to distinguish RTT and KEYFRAME_REQUEST from regular
+/// media data.
+///
 /// # Arguments
 /// * `data` - Raw packet bytes
 ///
 /// # Returns
 /// The classification of the packet
 pub fn classify_packet(data: &[u8]) -> PacketKind {
+    let packet_wrapper = match PacketWrapper::parse_from_bytes(data) {
```
darioalessandro: If the packet can't be parsed then throw it in the garbage

PR #760, opened 2026-03-14
perf 03: connection resilience - reconnection, re-election, heartbeats, datagrams

testradav:
## Summary
- Implements automatic reconnection with exponential backoff (1s to 16s cap, 10 max attempts)
- Fast-fail after 3 consecutive zero-connection attempts (stops retrying revoked sessions)
- RTT-based connection re-election: monitors at 1Hz, triggers when RTT exceeds 2x baseline for 5 consecutive samples
- Switches heartbeats from 1s polling to event-driven (5s keepalive + immediate on state change)
- Adds WebTransport datagram send path (send_packet_datagram) with automatic fallback to reliable stream for oversized packets
- Refactors ConnectionController to use Rc RefCell ConnectionManager directly with weak self-reference for reconnection callbacks
- Exports ReconnectionPhase enum (Idle/Reconnecting/Failed) for future UI status display

## Tests
- 25 unit tests: backoff calculation (4), RTT degradation (6), fast-fail (3), baseline RTT (4), connection state (4), re-election (4)

## PR Stack (replaces #755)
1. Foundation: constants + quality manager (#758, merged first)
2. Server hardening (merged second)
3. **This PR** - Connection resilience
4. Decoder: PLI keyframe requests + RED audio + visibility
5. Encoder adaptation + client wiring + UI integration

### DIFF
--- a/videocall-client/src/connection/connection_manager.rs
+++ b/videocall-client/src/connection/connection_manager.rs
@@ -1015,9 +1015,9 @@ impl ConnectionManager {
         false
     }
 
-    /// Begin a re-election: create new connections to all servers while keeping the
-    /// current active connection alive. Once election completes, switch to the new
-    /// best server seamlessly.
+    /// Begin a re-election: tear down existing connections, reset monitoring
+    /// state, and create fresh connections to all servers. There is a brief
+    /// disconnection period while the new election runs and selects a winner.
     pub fn start_reelection(&mut self) -> Result<()> {
         if self.reelection_in_progress {
             info!("Re-election already in progress, skipping");
@@ -1027,6 +1027,7 @@ impl ConnectionManager {
         info!("Starting connection quality re-election");
         self.reelection_in_progress = true;
         self.degradation_counter = 0;
+        self.baseline_rtt = None;
 
         // Close the old active connection before creating new ones.
         // `create_all_connections` reuses the same connection IDs (ws_0, wt_0, ...),
@@ -1050,6 +1051,12 @@ impl ConnectionManager {
         // Drain stale RTT responses from the previous connections.
         if let Ok(mut responses) = self.rtt_responses.try_borrow_mut() {
             responses.clear();
+        }
+
+        // Clear pending session IDs — connection IDs are reused (ws_0, wt_0, …)
+        // so stale entries could be incorrectly applied to new connections.
+        if let Ok(mut pending) = self.pending_session_ids.try_borrow_mut() {
+            pending.clear();
         }
 
         // Create fresh connections to all servers for testing.


darioalessandro:
@testradav why adding an exponential backoff for the reconnects? isnt it better to just retry asap forever? I do not think it is beneficial to:

1. take more between reconnects. 
2. give up.

Please explain me your goal with this change.

testradav:
addressed PR review

```  ┌─────┬──────────────────────────────────────────┬──────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────┐
  │  #  │                 Finding                  │                              Status                              │                     Action Needed?                      │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │     │ start_reelection() drops active          │ Already partially addressed — the initialize() fix on PR 762     │ Yes, but low priority — a true seamless handoff         │
  │ 1   │ connection before new election completes │ restructured this. On PR 760, the intentionally_disconnected     │ requires keeping the old connection alive until the new │
  │     │  (3s media gap)                          │ commit also improved re-election cleanup.                        │  one wins. This is a design change, not a bug fix.      │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 2   │ try_borrow_mut failure burns an attempt  │ Not fixed — line 863 falls through and the attempt counter       │ Yes — quick fix                                         │
  │     │ without retrying                         │ increments.                                                      │                                                         │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 3   │ DATAGRAM_MAX_SIZE duplicated across      │ Not fixed, cross-PR concern                                      │ No — documentation-level, not worth a code change now   │
  │     │ crates                                   │                                                                  │                                                         │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 4   │ Redundant protobuf parsing               │ Already fixed on PR 759 (e9a0b62). PR 760 has the same code but  │ No — already addressed                                  │
  │     │                                          │ will get the fix when rebased.                                   │                                                         │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 5   │ WebSocket transport has no congestion    │ Not a regression, just asymmetric                                │ No — document only                                      │
  │     │ tracking                                 │                                                                  │                                                         │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 6   │ warn! on every dropped CONGESTION packet │ Not fixed — line 349                                             │ Yes — quick fix                                         │
  │     │  (log flood risk)                        │                                                                  │                                                         │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ 7   │ 680-line design doc in diff              │ PR hygiene, not code                                             │ No                                                      │
  ├─────┼──────────────────────────────────────────┼──────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │     │ BITRATE_CHANGE_THRESHOLD unused in       │ The constant IS used — in camera_encoder.rs and                  │                                                         │
  │ 8   │ constants file                           │ screen_encoder.rs they define their own local copies. The one in │ Yes — quick fix (remove or wire up)                     │
  │     │                                          │  adaptive_quality_constants.rs is unused.                        │                                                         │
  └─────┴──────────────────────────────────────────┴──────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────┘
```

testradav:
> @testradav why adding an exponential backoff for the reconnects? isnt it better to just retry asap forever? I do not think it is beneficial to:
> 
> 1. take more between reconnects.
> 2. give up.
> 
> Please explain me your goal with this change.

good point.  here is a good compromise for both situations:

```
⏺ Fair pushback. Arguing for aggressive immediate retries without backoff or giving up. 

Good point about giving up. In a video call, if you disconnect, the user wants back in — period. RECONNECT_MAX_ATTEMPTS = 10 with exponential backoff means after ~30 seconds of failures, the client stops trying entirely. The user sees a dead screen with no recourse except refreshing. That's bad UX for a real-time communication app. The client should keep trying indefinitely (or until the user explicitly leaves).

However, removing backoff entirely is not optimal. Retrying "ASAP forever" with no backoff is a classic thundering herd / self-DDoS pattern. If the server goes down, all clients hammering it simultaneously at max rate will prevent recovery. A short backoff (even just 1-2 seconds, not the current ramp to 30s) is the right middle ground.

My recommendation:
  1. Remove RECONNECT_MAX_ATTEMPTS — never give up (retry indefinitely until user disconnects or succeeds)
  2. Keep backoff but cap it much lower — e.g., 500ms initial, 2s max instead of 1s initial / 30s max
  3. The intentionally_disconnected flag already handles the "user left" case cleanly

This gives the best of both worlds: fast reconnection without server-punishing retry storms
```

jboyd01:
1. Race window in start_reelection() — connections.clear() + rtt_measurements.clear() happens before create_all_connections() rebuilds them. If get_connection_state() fires in that gap while election_state is still Elected from the previous round, it emits a spurious Failed state to the UI. Narrow window but real.
  2. intentionally_disconnected never reset — Only initialized to false in new(). Fine today since a fresh ConnectionManager is created per session, but latent footgun if anyone reuses an
  instance.

  Minor:

 * consecutive_zero_connections check does another try_borrow that can fail silently, incorrectly incrementing the counter even if connections exist.
 * reconnect_retries_indefinitely test only asserts constants equal themselves — provides no real behavioral safety net.

testradav:
code changes addressed in https://github.com/security-union/videocall-rs/pull/760/commits/6a83208d757a1bb9241989378682ee73bafae82d :
- Stale docstring — updated to reflect actual teardown-first behavior                                                                                                             
- pending_session_ids not cleared — added clear before create_all_connections()
- baseline_rtt not reset — added reset with degradation_counter                                                                                                                   
- Borrow failure counted as "no connections" (item 9) — uses Option<bool>, borrow failure neither increments nor resets counter                                                   
- Duplicate attempt numbers in logs (item 8) — borrow conflicts now short-sleep 200ms and continue instead of falling through

testradav:
@darioalessandro ready for review

testradav:
PR #759 is a hard dependency — PRs #760 and #762 expect congestion feedback to work, but it's dead without PR 759's self-skip bypass fix. These must land together or #759 first.

PR #764, opened 2026-03-16
fix: stop media tracks after permission check to release camera/mic

jboyd01:
bug:  when you join a meeting, the camera light comes on and stays on regardless of if you are sharing your camera or not.

## Summary
- PR #745 split the single `getUserMedia({audio, video})` call into two separate calls for independent permission checking
- The returned `MediaStream` objects were discarded without stopping their tracks, causing the browser to keep camera/mic hardware active
- This manifested as the camera indicator light staying on after joining a meeting, even with video toggled off
- Fix: explicitly call `.stop()` on all `MediaStreamTrack`s before dropping the stream, so the browser releases the hardware immediately

## Test plan
- [ ] Join a meeting without toggling camera on — verify camera light does **not** stay on
- [ ] Join a meeting and toggle camera on — verify camera works normally
- [ ] Test on Chrome and Firefox
- [ ] Verify permission prompts still appear correctly for both audio and video


jboyd01:
I verified this fixes the issue.  You get two questions about allowing devices to be used (I **think** we should only get one with them combined?) but it definitely fixed the issue where before even though you aren't sharing a camera the camera light was on as though we were.

image
image

testradav (approved):
👍

PR #765, opened 2026-03-18
Allow admit message by anyone in the meeting vs just the host only

### LINKED ISSUE #742: Option to allow admit message by anyone in the meeting vs just the host
When attendees are in wait room, currently only the host is getting the "admit participant into the meeting..." message

Add option to meeting setup to toggle if "admit into meeting" should also be shown to anyone that has been already accepted in the meeting

palina-piliushyna:
Added support for a new meeting setting admitted_can_admit which allows already admitted participants to admit users from the waiting room when enabled.

Summary of changes:
- Added admitted_can_admit BOOLEAN NOT NULL DEFAULT FALSE to the meetings table
- Updated backend DB models and SQL queries to read/write the new field
- Updated meeting create/update flows to persist and return admitted_can_admit
- Updated waiting room authorization so:
  - host can always admit participants
  - admitted non-host participants can admit only when admitted_can_admit == true
- Updated shared request/response types
- Updated Dioxus frontend to expose the setting in meeting UI and use the new API field

Related context:
Previously only the host could receive/use the “admit participant into the meeting” flow. This change makes that behavior configurable per meeting instead of hardcoded.
## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [ ] Documentation updated
- [x] Before/After Images provided
Screenshot 2026-03-18 104157
Screenshot 2026-03-26 183150

photo_2026-03-26_18-38-54


jboyd01 (commented):
## Code Review

CI: All 10 checks pass. Well-structured feature — clean migration, solid backend auth, good test coverage.

### Moderate

**1. Two DB updates in `update_meeting` are not in a single transaction**

In `meeting-api/src/routes/meetings.rs`, when a request sends both `waiting_room_enabled` and `admitted_can_admit`, they execute as two separate DB operations. If the first succeeds and the second fails, the meeting is left partially updated. The UI sends both values when toggling waiting room off (setting `admitted_can_admit` to `false` at the same time). These should be wrapped in a single transaction.

**2. `admitted_can_admit` prop is never refreshed mid-meeting**

The `admitted_can_admit` value comes from the join/status response and is passed as a static prop. If the host toggles it OFF mid-meeting, already-admitted participants will still see the HostControls until they reload. The server-side auth will correctly reject the actions (so no security hole), but users will see confusing error messages instead of the controls disappearing gracefully.

### Minor

**3. No Yew UI parity** — The feature only exists in `dioxus-ui`. If Yew is being sunset, fine. If both UIs are in production, this is a gap.

### What looks good

- `require_host_or_can_admit()` is clean and checks admission status before the flag — correct auth layering
- Integration tests cover the three critical paths: can admit when enabled, cannot when disabled, waiting users cannot even when enabled
- Disabling waiting room automatically disables `admitted_can_admit` — good UX
- Migration has proper up/down

jboyd01 (changes requested):
Requesting changes for one item:

The `update_meeting` handler in `meeting-api/src/routes/meetings.rs` executes the `waiting_room_enabled` and `admitted_can_admit` updates as two separate DB operations. If the first succeeds and the second fails, the meeting is left partially updated. The UI sends both values when toggling the waiting room off.

Fix: wrap both updates in a single transaction, same pattern already used in `update_waiting_room_enabled`.

Everything else looks good — clean migration, solid auth logic in `require_host_or_can_admit()`, and strong integration test coverage.

testradav:
if I turn off `waiting room` does the other option gray out?

palina-piliushyna:
> if I turn off `waiting room` does the other option gray out?

image

testradav:
need to add option to the meeting options screen:

edit meeting button:
image


meeting options screen:
image

testradav:
Hi @piliushyna! This PR has a merge conflict in `dioxus-ui/src/components/host_controls.rs` that needs to be resolved before it can be merged.

To fix it, run:

```bash
git fetch upstream main
git merge upstream/main
```

The only conflict is in `dioxus-ui/src/components/host_controls.rs` (lines ~238-244). The resolution is straightforward — accept the incoming (main) version, which removes the unnecessary `.clone()` calls on the `error` signal (Dioxus `Signal` implements `Copy`, so `.clone()` is not needed):

```rust
// Replace the conflict block with:
let error_admit = error;
let error_reject = error;
```

Then commit and push:

```bash
git add dioxus-ui/src/components/host_controls.rs
git commit -m "Merge main: resolve host_controls.rs conflict"
git push
```

Let me know if you need any help!

testradav (changes requested):
resolve merge conflict

piliushyna:
device_integration tests were hanging because calls to get media devices sometimes never returned, so the test just got stuck until it timed out. I decided to add a real timeout so we don't wait forever. If getting real devices fails or times out, we fall back to mock devices so tests can still run and don't hang in CI

testradav (approved):

darioalessandro:
Now that yew-ui is gone, please remove the yew changes, that will also get the line count down XD

testradav:
close in lieu of https://github.com/security-union/videocall-rs/pull/854

PR #751, opened 2026-03-12
feat: build version info endpoints and diagnostics display

jboyd01:
## Summary

- Embeds git SHA (8-char), branch name, and build timestamp in all service binaries at compile time via `build.rs`
- Adds unauthenticated `/version` JSON endpoints on websocket relay, webtransport relay, and meeting-api
- Adds aggregated `/api/v1/versions` endpoint on meeting-api that fans out to peer services with 3s timeout
- Updates Dioxus diagnostics panel to show all component versions in a Build Info table
- Updates all 4 Dockerfiles, 6 GitHub Actions workflows, and 3 build scripts to pass build metadata

## Details

**Backend**: New `BuildInfo` typed struct ensures only expected fields are forwarded from peer services. Shared `reqwest::Client` stored in `AppState` for connection reuse. `SERVICE_VERSION_URLS` env var (comma-separated) configures which internal services to query.

**Frontend**: Diagnostics panel replaces static "Application Version" text with a 3-column table (Component, Commit, Branch) showing the UI's own build info plus all backend versions fetched from the aggregated endpoint.

**Build pipeline**: All Dockerfiles declare `ARG GIT_SHA/GIT_BRANCH/BUILD_TIMESTAMP` with "unknown" defaults. Build scripts support `REGISTRY` env var override. SHA standardized to 8 chars via `build.rs` truncation.

## Test plan

- [x] `cargo clippy -p videocall-api -p meeting-api` passes clean
- [x] Verify `/version` returns JSON on websocket relay (port 8080)
- [x] Verify `/version` returns JSON on webtransport health server (port 444)
- [x] Verify `/api/v1/versions` aggregates all configured services
- [x] Verify Dioxus diagnostics panel shows Build Info table with all components
- [x] Verify `docker build` with `--build-arg GIT_SHA=abc123` embeds the value
- [x] Verify local `cargo build` works without env vars set (falls back to git commands)

PR #768, opened 2026-03-23
feat: add SERVICE_VERSION_URLS to PR preview deploy workflow

jboyd01:
should have been done as part of https://github.com/security-union/videocall-rs/pull/751   This change sets the version urls so the UI diagnostics panel can show the version of each component.

## Summary
- Add `SERVICE_VERSION_URLS` env var to meeting-api in the PR preview deploy workflow
- Uses per-PR service names: `websocket-pr-N`, `dx-pr-N`, `ui-pr-N`
- No webtransport URL since it's disabled in preview environments
- Production already has this configured via global helm values

## Test plan
- [ ] Deploy a PR preview and verify `/api/v1/versions` endpoint returns service versions


testradav (approved):
👍

PR #769, opened 2026-03-24
bot: A/V sync, real VP9 encoding, conversation mode, RX diagnostics

jboyd01:
## Summary

Complete rework of the synthetic media bot for multi-participant conversation testing.

- **Dynamic N-participant mode** — single `generate-conversation-edge.py` run produces per-line WAV clips + manifest for 20 participants; `--users N` at runtime selects how many join with no silent gaps
- **On-the-fly EKG video** — Rust renderer generates 1280x720 waveform frames in the video loop (< 1ms per frame), no pre-generation or startup delay
- **Real VP9 encoding** via libvpx with forced keyframes at loop restart and every 5s
- **Microsecond-precision A/V timing** — both producers derive position from a shared `Instant` clock with µs math, preventing the ~843ms/84s progressive drift caused by integer truncation at ms resolution
- **OS threads for media producers** — `std::thread::spawn` + `std::thread::sleep` avoids tokio scheduler starvation under CPU-bound VP9/Opus encoding
- **WebSocket transport** — new WS client with JWT auth and proper Pong responses (was WebTransport only)
- **RX quality diagnostics** — per-sender jitter, sequence gaps, A/V sync delta, keyframe count reported every 10s
- **Simplified config** — `config.yaml.template` + CLI `--config`/`--users` flags; `config*.yaml` gitignored to prevent secret leaks
- **Pre-allocated frame buffers** — reusable ImageBuffer + I420 buffer eliminates ~4.1 MB/frame heap churn at scale
- **Single-threaded VP9 encode** — `g_threads=1` (was 2), halves libvpx thread count; well within 66ms budget at `cpu_used=5`

### Architecture

```
generate-conversation-edge.py (one-time)
  └─> conversation/manifest.yaml + lines/line_NNN.wav

bot --config config.yaml --users N (runtime)
  ├─ Reads manifest, takes first N participants
  ├─ Filters conversation lines to active speakers
  ├─ Stitches per-participant audio in memory (speech + silence)
  ├─ Computes loop_duration from stitched timeline
  └─ Per participant:
       ├─ AudioProducer (OS thread, 50fps Opus, 20ms packets)
       └─ VideoProducer (OS thread, 15fps VP9, on-the-fly EKG render)
```
## Checkout out the updated README

[README.md](https://github.com/jboyd01/videocall-rs/blob/bf9569f5ae6348225920b2a709ff86d3f785c289/bot/README.md)

## Test plan

- [x] Generate conversation: `python3 generate-conversation-edge.py`
- [x] Copy template: `cp config.yaml.template config-myenv.yaml` and fill in values
- [x] Run with WebSocket: `bot --config config-myenv.yaml --users 5`
- [x] Run with WebTransport: update transport in config, run with `--users 5`
- [x] Verify A/V sync stays tight through full conversation loop (no progressive drift)
- [x] Verify EKG animation matches audio energy (speaking = waveform, silent = flat pulse)
- [x] Scale test: `--users 20` on 4+ core instance
- [x] Verify RX STATS lines appear every 10s in logs
- [x] Verify keyframe forcing at loop boundary

## Demo (make sure you enable sound in the video player)

https://github.com/user-attachments/assets/4f599053-4237-4c8d-93ca-b167b214f1d8


testradav:
should have some of the bots speaking over each other to mimic real humans.  they are too respectful

testradav (approved):
is this part of e2e test, or a manually kicked off test.
are the bot voices going through client or injected in server
trying to understand how this will be leveraged

jboyd01:
> is this part of e2e test, or a manually kicked off test. are the bot voices going through client or injected in server trying to understand how this will be leveraged

Manually run.  Instead of asking people to join your meeting, you can build and run the bot to see how it works.  Helpful if you are fine tuning the glow around the speaker, ensuring the glow doesn't bleed through things it shouldn't (like the settings or diagnostics panel) etc.

videocall-client/src/scripts/pcmPlayerWorker.js
```diff
@@ -32,7 +32,8 @@ class UltraFastPCMBuffer {
         console.log(`Ultra-fast PCM buffer: ${maxSamples} samples (~${(maxSamples/48000*1000).toFixed(1)}ms), ${channels} channels`);
     }
     
-    // Drop the oldest samples to make room (advance read pointer)
+    // Drop the oldest samples to make room (advance read pointer).
```
darioalessandro, line 35: @jboyd01 this is unrelated to the bot

bot/src/video_encoder.rs
```diff
@@ -92,14 +92,14 @@ impl VideoEncoderBuilder {
...
-        cfg.g_threads = 2;
+        cfg.g_threads = 1;
         cfg.g_lag_in_frames = 1;
         cfg.g_error_resilient = VPX_ERROR_RESILIENT_DEFAULT;
         cfg.g_pass = vpx_enc_pass::VPX_RC_ONE_PASS;
         cfg.g_profile = self.profile;
         cfg.rc_end_usage = vpx_rc_mode::VPX_VBR;
         cfg.kf_max_dist = 150;
-        cfg.kf_min_dist = 150;
+        cfg.kf_min_dist = 0;
```
darioalessandro, line 102: This defines the min number of frames between keyframes, I think setting it to 0 tells the encoder to insert a frame whenever it feels it needs one, why do we need this change?

darioalessandro (changes requested):
This PR looks fascinating, I'll review tonight in detail, I love the text synth so that we do not have to include all those wav files, do you do something similar for video?

PR #791, opened 2026-03-26
Fix: entry exit messages

testradav:
- **Built on top of PR #778**

- Fix false join/leave toast storm caused by aggressive heartbeat timeouts and reconnection replays

## Changes:
1. Increased CLIENT_TIMEOUT from 10s to 30s (tolerate 3 missed heartbeats instead of 1)
2. Increased QUIC_MAX_IDLE_TIMEOUT_SECS from 5 to 15-30
3. Added a server-side reconnection grace period — if the same user_id reconnects within N seconds, suppress the PARTICIPANT_LEFT/JOINED events
4. Client-side: suppressed toasts for already-known participants — if participant X is already in the peer list, don't show a "joined" toast
5. Increased the client dedup window from 5s to 30s+ to survive reconnection backoff

also fix adaptive quality oscillation and peer removal/rejoin https://github.com/security-union/videocall-rs/pull/791#issuecomment-4142567577

```
  ┌─────┬────────────────────────────────────────────────┬─────────────┐                                                                                   
  │  #  │                      Fix                       │   Status    │                                                                                 
  ├─────┼────────────────────────────────────────────────┼─────────────┤                                                                                   
  │ 1   │ Increase CLIENT_TIMEOUT 10s → 30s              │ ✅ Done     │                                                                                   
  ├─────┼────────────────────────────────────────────────┼─────────────┤                                                                                   
  │ 2   │ Increase QUIC_MAX_IDLE_TIMEOUT_SECS 5 → 30     │ ✅ Done     │                                                                                   
  ├─────┼────────────────────────────────────────────────┼─────────────┤                                                                                   
  │ 3   │ Server-side reconnection grace period (2s)     │ ✅ Done     │                                                                                   
  ├─────┼────────────────────────────────────────────────┼─────────────┤                                                                                   
  │ 4   │ Suppress toasts for already-known participants │ ✅ Done     │                                          
  ├─────┼────────────────────────────────────────────────┼─────────────┤                                                                                   
  │ 5   │ Increase client dedup window 5s → 30s          │ ✅ Done     │
  └─────┴────────────────────────────────────────────────┴─────────────┘         
```

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [x] No tests needed

## Other
- [ ] Documentation updated


anhelinakrutaya-ops:
@testradav
 I checked these changes, but they don’t solve the user reconnect issue. The user is still leaving and then joining again even when they didn’t intend to. Also, the session keeps restarting, and this happens several times per minute. This frequent reconnect mostly occurs when webtransport is active. With websocket, reconnects also happen, but not as often, which is probably why they're less noticeable.

Alena-Melnikava:
From my side I can add that the most important problem that connection is really recreated. When connection is recreated we reset all activity which was done by user: for example, if user activate share screen or microphone usage, after reconnection his screen share and microphone will be swithed off. Therefore, it is very important to eliminate unnecessary reconnections, and not just hide them.

testradav:
Fixing issues identified:

1. Adaptive quality oscillation on startup — fps_ratio=0.00 immediately triggers aggressive tier downgrades (high → low → minimal) before the encoder has produced any frames, then rapidly upgrades back up. This causes the aspect ratio glitch (640x480 → 640x360 → 426x240 → back up).
2. Peer removal/re-creation cycle — "Peer removed: 10736890868112722680" followed by "New user joined: 10736890868112722680" with the same session ID. This is NOT a server reconnect (session ID would change). Something client-side is removing the peer, then re-adding it when media packets keep arriving.


## Root causes:

### Bug 1: Peer removed/re-added every ~100 seconds

- Root cause: Two unsynchronized 5-second setInterval timers — one sends heartbeats, the other checks heartbeat_count. When the check timer fires just before the next heartbeat arrives, heartbeat_count == 0 → peer removed. Then the next media packet triggers ensure_peer() → peer re-created.             

- Worse: only HEARTBEAT packets increment the counter. A peer actively streaming 30fps video but whose heartbeat happens to arrive 1ms late is considered "dead."

### Bug 2: Adaptive quality oscillation on startup                                                                   

- Root cause: No warmup period. The quality manager starts checking fps_ratio immediately after the encoder is created, before it produces any frames. fps_ratio=0.00 → immediate downgrade (high → low → minimal). last_transition_time_ms starts at 0.0, so transitions are allowed immediately. Once frames flow, it climbs back up through every tier, each time reconfiguring the encoder and changing dimensions.

### Bug 3: User appearing twice

- Side effect of Bug 1 — the peer is removed and re-created, and if a real reconnection overlaps with this cycle, two session IDs for the same user coexist briefly.

## Fixes:                                                                                                  
                                                   
### Bug 1: 

Count ANY inbound media packet (audio/video/screen) toward liveness, not just HEARTBEAT. A peer actively sending video is obviously alive. Also increase the tolerance from 0 to at least 1 missed check before removal.

### Bug 2: 

Add a warmup grace period — skip quality downgrades for the first few seconds after encoder start, giving it time to produce frames.

testradav:
build-and-deploy

testradav:
tested successfully with 4 participants (websocket only due to sandbox pr) https://pr3.sandbox.videocall.rs/meeting/test_entry

anhelinakrutaya-ops:
> tested successfully with 4 participants (websocket only due to sandbox pr) https://pr3.sandbox.videocall.rs/meeting/test_entry

Tested with webtransport - the reconnect issue still occurs. I tested with two users, and it now appears that a new phantom user connects and disconnects each time, while the other two remain unchanged.

testradav:
fixing webtransport (replicated the problem in my local env)...

## Root Causes

### Client-side (3 fixes already applied by frontend agent):                                                                                                               
                                                                                                                                                                         
1. "Stream is locked" — webtransport.rs was calling transport.close() when a datagram write found the stream locked (benign concurrency between heartbeat/RTT/media).  
Fix: drop the packet silently instead of killing the connection
2. RTT threshold too aggressive — Baseline 0.9ms × 2.0 = 1.8ms threshold; normal jitter exceeds this. Fix: multiplier raised to 3.0 and a 50ms minimum floor added so localhost never false-triggers.
3. Re-election gap — Old connection was closed before new one was ready, causing a visible disconnect. Fix: old connection kept alive as fallback during re-election, only dropped after new winner is established.
 
### Server-side (plan proposed, not yet implemented):                                                                                                                      
                                                                                                                                                                         
1. Ghost PARTICIPANT_JOINED — Server broadcasts join events when JoinRoom is called, even for Testing connections. Both WT and WS test connections trigger this.
2.  Ghost PARTICIPANT_LEFT — The connection_states map is cleaned up in Disconnect before the grace period check in ExecutePendingDeparture, so the Testing guard is defeated.

.claude/settings.local.json
```diff
@@ -51,7 +51,9 @@
       "WebFetch(domain:raw.githubusercontent.com)",
       "WebFetch(domain:docs.rs)",
       "WebFetch(domain:crates.io)",
-      "Bash(cargo clippy:*)"
+      "Bash(cargo clippy:*)",
+      "Bash(openssl x509:*)",
```
testradav, line 55: so that I could run webtransport in localhost

CLAUDE.md
```diff
@@ -49,6 +49,16 @@ Run agents in parallel when tasks are independent. Always run `code-reviewer` af
...
+- **Consider all transport modes.** Changes to shared connection logic must be validated against both WebTransport and WebSocket paths. A fix for one transport must not introduce regressions in the other.
+- **Consider real-world networks.** Thresholds, timeouts, and retry logic must account for high-latency links (200ms+), packet loss, jitter, and mobile networks — not just localhost. Hardcoded values that only work on fast local connections are bugs.
+- **Consider scale.** Meetings may have many participants. Events that fire per-connection (not per-user) can cause O(n) storms during reconnection waves. Session management, NATS publishing, and UI re-renders must all be evaluated for fan-out cost.
+- **Consider the server as part of the system.** Client-side fixes that rely on server behavior (e.g., session lifecycle, event broadcasting) must verify the server actually upholds those assumptions, and vice versa. Cross-cutting changes require both frontend and backend agents.
+
```
testradav, line 61: so that changes are done in the context of the purpose of the app instead of nit picking in isolated code spaces

testradav:
#793 merged. closing this PR because 793 contains these changes

PR #802, opened 2026-03-30
feat: add SQLite as optional database backend for meeting-api

### LINKED ISSUE #795: Add SQLite support as an alternative to PostgreSQL for the database backend
motivation: for self-hosted / small-to-medium deployments, SQLite eliminates the need for a separate database server, making deployment on tiny  servers much simpler. PostgreSQL would remain the default for larger-scale setups.

from @holg https://github.com/holg

holg:
Add compile-time feature flag to select between PostgreSQL (default) and SQLite for the meeting-api database backend. This enables simpler deployment on small servers without requiring a separate PostgreSQL instance.

- Add `postgres` (default) and `sqlite` feature flags to Cargo.toml
- Restructure db/ into postgres/ and sqlite/ backend modules
- Introduce DbPool type alias for backend-agnostic pool references
- Adapt all SQL queries for SQLite dialect (parameter binding, datetime functions, no FOR UPDATE row locking)
- Add SQLite migration with consolidated schema
- Configure WAL mode and busy timeout for SQLite connections
- Update state.rs, main.rs, auth.rs and integration tests to use DbPool

Build with SQLite: cargo build -p meeting-api --no-default-features --features sqlite

_Please include a summary of the changes and related context._

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Test plan
  - [x] `cargo fmt -p meeting-api -- --check` — pass
  - [x] `cargo clippy -p meeting-api -- -D warnings` (postgres) — pass, 0 warnings
  - [x] `cargo clippy -p meeting-api --no-default-features --features sqlite -- -D warnings` — pass, 0 warnings
  - [x] `cargo check -p meeting-api` (postgres) — pass
  - [x] `cargo check -p meeting-api --no-default-features --features sqlite` — pass
  - [x] `cargo test -p meeting-api --lib` — 88/88 pass

## Testing
- [x] No tests needed

## Other
- [ ] Documentation updated

Fixes issue #795


darioalessandro:
Hey @holg thanks for your contribution!! looking forward to getting this in, give me a couple of days to get to it

meeting-api/Cargo.toml
```diff
@@ -17,6 +17,11 @@ path = "src/lib.rs"
 name = "meeting-api"
 path = "src/main.rs"
 
+[features]
```
darioalessandro, line 20: @holg nice!! yes this is the pattern that we want!

meeting-api/src/state.rs
```diff
@@ -16,14 +16,14 @@
...
 use crate::oauth::JwksCache;
-use sqlx::PgPool;
 
 /// Application state shared across all request handlers.
 #[derive(Clone)]
 pub struct AppState {
-    /// PostgreSQL connection pool.
-    pub db: PgPool,
+    /// Database connection pool (PostgreSQL or SQLite depending on feature flag).
+    pub db: DbPool,
```
darioalessandro, line 26: generic DbPool! perfect

meeting-api/src/db/sqlite/meetings.rs
```diff
@@ -0,0 +1,210 @@
...
+    room_id: &str,
+    creator_id: &str,
+    password_hash: Option<&str>,
+    attendees: &JsonValue,
+    waiting_room_enabled: bool,
+) -> Result<MeetingRow, sqlx::Error> {
+    sqlx::query_as::<_, MeetingRow>(
+        r#"
+        INSERT INTO meetings (room_id, creator_id, started_at, password_hash, state, attendees, waiting_room_enabled)
+        VALUES (?1, ?2, datetime('now'), ?3, 'idle', ?4, ?5)
```
darioalessandro: Fun discovery while testing this: sqlx's SQLite driver parses `$1, $2, ...` natively (out-of-order and repeated too), and SQLite itself accepts them. So the `?N` rewrite isn't needed — this whole module can share the postgres query strings verbatim. That's 25 of the 48 divergent lines gone, and the main thing standing between this PR and a single shared `db/` module.

meeting-api/src/db/sqlite/participants.rs
```diff
@@ -0,0 +1,255 @@
...
+///
+/// Returns `(auto_admitted, ParticipantRow, waiting_room_enabled)` where `auto_admitted`
+/// is `true` when the participant was immediately admitted (waiting room disabled).
+pub async fn join_attendee(
+    pool: &DbPool,
+    meeting_id: i32,
+    user_id: &str,
+    display_name: Option<&str>,
+) -> Result<(bool, ParticipantRow, bool), sqlx::Error> {
+    let mut tx = pool.begin().await?;
```
darioalessandro: The doc comment above says IMMEDIATE, but `pool.begin()` issues `BEGIN` (DEFERRED) — a read snapshot with no write lock. We could reproduce a race where a join interleaved with a waiting-room toggle strands a participant in `'waiting'` forever. Fix is small:
```rust
let mut tx = pool.begin_with("BEGIN IMMEDIATE").await?;
```
It needs to go on **both** this path and `update_waiting_room_enabled` in `meetings.rs` (locking only one side moves the race), plus a small retry on `SQLITE_BUSY` — `busy_timeout` doesn't cover `SQLITE_BUSY_SNAPSHOT`.

meeting-api/src/main.rs
```diff
@@ -38,13 +37,38 @@ async fn main() {
...
+    #[cfg(feature = "sqlite")]
+    let pool = {
+        use sqlx::sqlite::SqlitePoolOptions;
+        let pool = SqlitePoolOptions::new()
+            .max_connections(5)
+            .connect(&config.database_url)
+            .await
+            .expect("failed to connect to SQLite");
+        // Enable WAL mode for better concurrent read performance.
+        sqlx::query("PRAGMA journal_mode=WAL")
```
darioalessandro: These PRAGMAs run on one arbitrary pooled connection — `busy_timeout` is per-connection, so the other 4 keep the default of 0, and `foreign_keys` is never set at all, which silently disables every `ON DELETE CASCADE` in the schema. Moving them onto the connect options covers every connection:
```rust
SqliteConnectOptions::from_str(&config.database_url)?
    .journal_mode(SqliteJournalMode::Wal)
    .busy_timeout(Duration::from_secs(5))
    .foreign_keys(true)
```

meeting-api/Cargo.toml
```diff
@@ -17,6 +17,11 @@ path = "src/lib.rs"
...
+default = ["postgres"]
+postgres = ["sqlx/postgres"]
+sqlite = ["sqlx/sqlite"]
```
darioalessandro, line 23: `cargo build --all-features` currently enables both backends and defines `DbPool` twice. Three lines in `db/mod.rs` turn that into a clear error:
```rust
#[cfg(all(feature = "postgres", feature = "sqlite"))]
compile_error!("features `postgres` and `sqlite` are mutually exclusive");
#[cfg(not(any(feature = "postgres", feature = "sqlite")))]
compile_error!("enable exactly one of `postgres` or `sqlite`");
```

dbmate/sqlite/migrations/20220807000000_initial_schema.sql
```diff
@@ -0,0 +1,93 @@
...
+CREATE UNIQUE INDEX IF NOT EXISTS idx_meetings_room_id_unique_active
+    ON meetings(room_id) WHERE deleted_at IS NULL;
+
+CREATE INDEX IF NOT EXISTS idx_meetings_room_id ON meetings(room_id);
+CREATE INDEX IF NOT EXISTS idx_meetings_creator_id ON meetings(creator_id);
+CREATE INDEX IF NOT EXISTS idx_meetings_state ON meetings(state);
+
+-- Auto-update updated_at on meetings
+CREATE TRIGGER IF NOT EXISTS update_meetings_updated_at
+    AFTER UPDATE ON meetings
```
darioalessandro: SQLite computes `RETURNING` *before* AFTER-triggers fire, so every `UPDATE ... RETURNING` gets the stale `updated_at` (we verified this by test). Simplest fix: drop both triggers and set `updated_at` explicitly in the UPDATE statements — on postgres that write is harmless since it matches what the trigger produces.

dbmate/sqlite/migrations/20220807000000_initial_schema.sql
```diff
@@ -0,0 +1,93 @@
...
+    last_login TEXT DEFAULT (datetime('now'))
+);
+
+-- Meetings
+CREATE TABLE IF NOT EXISTS meetings (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    room_id TEXT NOT NULL,
+    started_at TEXT NOT NULL,
+    ended_at TEXT,
+    created_at TEXT NOT NULL DEFAULT (datetime('now')),
```
darioalessandro: Heads-up on format: `datetime('now')` stores `'2026-07-21 04:39:04'` while a Rust-bound `DateTime<Utc>` stores RFC3339 with a `T` — mixed in one column they sort wrong lexicographically. Using `strftime('%Y-%m-%dT%H:%M:%f+00:00','now')` for the DEFAULTs keeps everything in one sortable format. Also worth adding while here: `CHECK (json_type(attendees) = 'array' AND json_array_length(attendees) <= 100)` (matches postgres's `chk_attendees_max_100` — the `json_type` part matters, `json_array_length` alone accepts `{"a":1}`) and `CHECK (length(col) <= 255)` where the `VARCHAR(255)`s were load-bearing, since SQLite ignores `VARCHAR(n)`.

dbmate/sqlite/.dbmate.yml
```diff
@@ -0,0 +1,3 @@
+db: "${DATABASE_URL}"
+migrations_dir: "./sqlite/migrations"
```
darioalessandro: dbmate resolves this relative to the config file, so it points at `dbmate/sqlite/sqlite/migrations`, which doesn't exist. Easiest fix: mirror the existing layout (`dbmate/sqlite/db/migrations`) and delete this file — dbmate's defaults then just work.

darioalessandro (commented):
Thanks @holg — we want this feature, and your PR already proved the hard part: `RETURNING`, `ON CONFLICT ... DO UPDATE`, and the partial unique index all work on SQLite as-is. 🎉

The big opportunity: the two backend modules differ by only ~48 lines out of ~1,125, and the inline comments below show how almost all of those can disappear — sqlx's SQLite driver accepts `$N` placeholders natively, so `db/postgres/` and `db/sqlite/` can collapse into one shared module, leaving just a tiny cfg'd helper for the clock and one for the `FOR UPDATE` lock. That would make this PR dramatically smaller and give every future query both backends for free.

Happy to help along the way — details are in the inline comments.

holg:
Thanks for the detailed review @darioalessandro — all inline comments are addressed in 2b269b30, and the `$N` discovery did exactly what you predicted: `db/postgres/` and `db/sqlite/` are gone, replaced by one shared `db/{meetings,participants,oauth}.rs` using the postgres query strings verbatim (net **−300 lines** on top of the previous state).

The only backend-specific pieces now live in `db/mod.rs`:
- `SQL_NOW` — `NOW()` vs `strftime('%Y-%m-%dT%H:%M:%f+00:00','now')` (RFC3339 with ms, so DEFAULTs, SQL-side writes, and sqlx-bound `DateTime<Utc>` all sort consistently, per your format comment)
- `FOR_UPDATE` — `" FOR UPDATE"` vs `""`
- `begin_write_tx` — plain `BEGIN` vs `BEGIN IMMEDIATE` (via `Pool::begin_with`)
- `with_busy_retry` — retries on SQLITE_BUSY primary code 5 (covers `SQLITE_BUSY_SNAPSHOT` 517) with short backoff; compile-time pass-through on postgres

Point-by-point:
- **BEGIN IMMEDIATE race** — both `join_attendee` and `update_waiting_room_enabled` now use `begin_write_tx` + `with_busy_retry`, so the join/toggle race you reproduced is closed on both sides. The 16 `waiting_room_toggle` tests pass against SQLite.
- **Per-connection PRAGMAs** — one-shot PRAGMAs replaced with `SqliteConnectOptions` (`journal_mode(Wal)`, `busy_timeout(5s)`, `foreign_keys(true)`), so all pooled connections get them — you were right that FK enforcement (and every `ON DELETE CASCADE`) was silently off before.
- **compile_error! guards** — added both; `--all-features` and no-backend builds now fail with a clear message.
- **Triggers vs RETURNING** — both AFTER UPDATE triggers dropped (verified the stale-`updated_at` behavior empirically); every UPDATE on `meetings`/`meeting_participants` sets `updated_at` explicitly — harmless double-write next to the pg trigger.
- **Datetime format** — all DEFAULTs use the strftime RFC3339 form you suggested.
- **CHECKs** — added `CHECK (json_type(attendees) = 'array' AND json_array_length(attendees) <= 100)` and `length() <= 255` on the formerly load-bearing VARCHAR(255) columns (room_id, creator_id, host_display_name, user_id, display_name, users.email).
- **dbmate layout** — migration moved to `dbmate/sqlite/db/migrations/`, `.dbmate.yml` deleted; `cd dbmate/sqlite && dbmate up` works with dbmate defaults and `db/schema.sql` is committed like the pg one.

Copilot's catches are in too: the stale `INSERT ... ON CONFLICT` doc comment is fixed (it's a plain INSERT; duplicates fail via the partial unique index), `oauth_requests.csrf_state` is now `UNIQUE` (note: pg doesn't have this constraint — happy to drop it or add a matching pg migration, your call), and there's a new `meeting-api-sqlite` CI job (dbmate v2.34.1-migrated file DB, clippy + the full integration suite with `--features sqlite`) plus `meeting-api/**` added to the workflow path filter and a `make tests_sqlite_run` target for local runs.

Verified: `cargo fmt --check`, clippy `-D warnings` on both feature sets, 88 lib tests on both backends, and the full 58-test integration suite green against a migrated SQLite db; both compile_error guards fire as intended.

darioalessandro:
Thanks for your contribution @holg !! I am looking at it right now 😄

PR #813, opened 2026-03-31
fix: track session instance id on reconnection to avoid phantom sessions 

### LINKED ISSUE #805: Perf: video adaptation should start from middle ground quality rather than low end
When a peer joins a meeting, the quality adaptation starts by providing the video feed using the lowest resolution and ramps up from there if the network bandwidth allows it.  Instead of always starting from the low end, the app should start from a middle-ground resolution deemed acceptable and work up or down from there.

### LINKED ISSUE #808: Perf:  content sharing needs to start from middle ground quality
when sharing, content is not viewed legibly because of perf. optimization

similar to #805

### LINKED ISSUE #811: perf:  entry/exit messages with some peers
# Issue 1 - Peers Appearing to Leave and Rejoin (Alena):
- alena.melnikava@hcl-software.com appears with 3 different session IDs: 10944137450972219485, 9561639356972527749, 12525607367443106996                               
- The 2nd-to-3rd transition is near-instant (~276ms between remove and new join), which is what triggers the "left and rejoined" toast                                 
- This suggests her client's WebSocket is dropping and reconnecting, generating new sessions each time   


**Root cause:** Race condition in session lifecycle + grace period too short

There are two scenarios causing this:                                                                                                                                  
                                                                                                                                      
#### Scenario A — Silent network drop (most common for far peers):                                                                                                          
The client detects the failure and reconnects before the server's 30-second heartbeat timeout fires on the old session. The new session's JoinRoom arrives while the old session is still alive on the server. The pending_departures grace period mechanism is never triggered because the old session hasn't disconnected yet. 
Result: PARTICIPANT_JOINED (new) broadcasts immediately, then 5-30s later PARTICIPANT_LEFT (old) broadcasts — peers see a confusing join-then-leave.
  - actix-api/src/actors/chat_server.rs — JoinRoom handler doesn't check for existing sessions with the same (room, user_id) when the old session is still alive         
  - actix-api/src/constants.rs:26 — CLIENT_TIMEOUT = 30s means slow detection of dead sessions
                                                                                                                                                                         
#### Scenario B — Clean disconnect, slow reconnection:                                                                                                                      
The 2-second RECONNECT_GRACE_PERIOD (actix-api/src/constants.rs:32) is too short for the client's reconnection cycle, which includes 500ms backoff + election period + 500ms post-election + handshake time — routinely exceeding 2 seconds. The grace period expires, PARTICIPANT_LEFT broadcasts, then the reconnection completes and PARTICIPANT_JOINED broadcasts.                                                                                                      


## Fix needed (server): 

In JoinRoom, detect existing room_members with the same user_id but different session_id — evict the old session silently and suppress both events.

### LINKED ISSUE #814: WebTransport: per-packet uni streams cause continuous keyframe request storm
## Summary

WebTransport sends each video packet on a **new unidirectional QUIC stream** (`send_unidirectional_stream`). QUIC guarantees ordering *within* a stream but **not across streams**, so video packets routinely arrive out of order. The sequence gap detector interprets this as packet loss and fires continuous keyframe requests (PLI), creating a bandwidth death spiral that can freeze video entirely.

This does **not** affect WebSocket connections, which use a single TCP stream with strict FIFO ordering.

### Transport layer (videocall-client/src/connection/webtransport.rs:103-104)
```rust
fn send_bytes(&self, bytes: Vec<u8>) {
    WebTransportTask::send_unidirectional_stream(self.transport.clone(), bytes);
}
```
Every "reliable" video packet opens a new uni stream. Two consecutive packets (seq 99, seq 100) sent on separate streams can arrive in any order.

### Gap detector (videocall-client/src/decode/peer_decode_manager.rs:491-556)
```rust
fn track_sequence(&mut self, media_type: MediaType, packet: &MediaPacket) -> Option<MediaType> {
    // ...
    if seq > prev + 1 {
        // Gap detected — record the time
        if gap_at.is_none() {
            *gap_at = Some(now);
        }
    }
    // Gap only clears when a keyframe arrives (frame_type == "key")
    if frame_type_str == "key" {
        *gap_at = None;
    }
    // After KEYFRAME_REQUEST_TIMEOUT_MS (1s), fire requests every 500ms
    if elapsed_since_gap >= KEYFRAME_REQUEST_TIMEOUT_MS
        && elapsed_since_last_req >= KEYFRAME_REQUEST_MIN_INTERVAL_MS { ... }
}
```

- Assumes ordered delivery (designed for WebSocket)
- Out-of-order arrival (`seq > last_seq + 1`) is treated as a gap
- Gap only clears on keyframe arrival, NOT when the "missing" packet shows up
- Once triggered, fires a keyframe request every 500ms until a keyframe arrives
- Keyframes only occur naturally every ~150 frames (5s at 30fps), so there are many opportunities for new out-of-order packets to re-trigger gaps between keyframes

## The Death Spiral

1. Normal out-of-order delivery on WebTransport triggers gap detection
2. After 1 second, keyframe requests start firing every 500ms
3. Both browsers enter this state simultaneously (mutual PLI storm)
4. Each keyframe request forces the remote encoder to produce a keyframe (5-10x larger than a delta frame)
5. Continuous keyframe generation consumes bandwidth — `bitrate_ratio` drops to ~0.39
6. Adaptive quality manager steps down: high → medium → low → minimal (426x240 @ 10fps)
7. Audio follows: high → medium → low → emergency (16kbps)
8. Eventually one browser's encoder can't keep up at all — stops producing decodable output
9. The other browser floods with unanswered keyframe requests indefinitely (671 consecutive requests observed with no response)
10. Video freezes permanently — no recovery without page reload

### Log 1 (two browsers, stable connection):
- Continuous bidirectional keyframe exchange throughout entire session
- Both "Received KEYFRAME_REQUEST" and "Sending KEYFRAME_REQUEST" every ~30-50 frames
- 159 forced keyframes observed in a single session
- Adaptive quality cycling: drops when bots join, recovers after — but PLI storm persists even with just 2 participants

### Log 2 (two browsers, one dies):
- Same continuous PLI storm initially
- Line 707: `bitrate_ratio=0.39` — only achieving 39% of target bitrate
- Rapid step-down: high → medium → low → minimal → audio emergency
- After line 760: "Received KEYFRAME_REQUEST" messages **completely stop** — the other browser has gone silent
- 671 consecutive unanswered "Sending KEYFRAME_REQUEST" messages (lines 760-1294)
- `datagram dropped (stream busy)` errors scattered throughout
- **No recovery** — log ends still spamming keyframe requests

### Option A: Fix the transport (preferred)
Use a **single persistent stream** for video instead of opening a new uni stream per packet. This preserves ordering and makes the gap detector work correctly.

### Option B: Fix the gap detector
Make `track_sequence()` tolerant of out-of-order delivery:
- Use a **reorder buffer / sliding window** instead of simple `seq > last_seq + 1`
- Clear the gap when the "missing" packet eventually arrives, not just on keyframes
- Add a **maximum outstanding requests** limit (e.g., stop after 5-10 unanswered requests)

### Option C: Both
Fix the transport for correctness, and make the gap detector robust for defense in depth.

### Additional mitigations (regardless of fix choice):
- **Rate-limit keyframe requests more aggressively** — 500ms is too frequent when keyframes are 5-10x normal frame size
- **Add a give-up threshold** — after N unanswered requests, show "video unavailable" instead of flooding
- **Exponential backoff** on keyframe requests instead of fixed interval

## Affected Code

| File | Lines | Role |
|------|-------|------|
| `videocall-client/src/connection/webtransport.rs` | 103-104 | Opens new uni stream per packet |
| `videocall-client/src/connection/websocket.rs` | 56-58 | Single stream (not affected) |
| `videocall-client/src/decode/peer_decode_manager.rs` | 491-556 | Gap detection + keyframe request logic |
| `videocall-client/src/adaptive_quality_constants.rs` | 308-312 | `KEYFRAME_REQUEST_TIMEOUT_MS=1000`, `MIN_INTERVAL_MS=500` |

## Environment
- my "ascend" deployment (WebTransport over internet, not localhost)
- Chrome 146
- Two browser participants (no bots)
- videocall-client from `vcprobe` branch

testradav:
Solution:  track client session instance by issuing the session id on a reconnection when a client drops momentarily.  using the session id for the client instance allows for multiple instances connected to the same meeting using the same user id.  using the session id avoids collision or mixing the sessions basing them solely on the user id.

# Fix: Spurious peer rejoin notifications

## Problem
During meetings, peers (especially those geographically far from the server) appeared to leave and immediately rejoin. This caused confusing toast notifications for other participants.

**Root cause:** When a client's connection dropped, it reconnected and obtained a new server session *before* the server's 30-second heartbeat timeout detected the old session was dead. The server had two live sessions for the same user with no way to correlate them, so it broadcast `PARTICIPANT_JOINED` (new session) followed later by `PARTICIPANT_LEFT` (old session timeout) — appearing as a spurious rejoin.

The existing 2-second reconnection grace period (`pending_departures`) only worked when the server detected the disconnect *before* the client reconnected. In the common case for high-latency links, the client reconnects faster than the server detects the failure, so the grace period was never triggered.

## Solution
When reconnecting, the client now tells the server which session it is replacing.

**Client** (`connection_manager.rs`): On reconnection, appends `&previous_session_id=<old_id>` to lobby URLs. The session ID is already known from the `SESSION_ASSIGNED` message received during the previous connection. First-time connections send no parameter.

**Server** (`chat_server.rs` JoinRoom handler): When a new session provides `previous_session_id`, the server:
1. Looks it up directly in `room_members` (no extra data structures needed)
2. Verifies it belongs to the same `user_id` (JWT-derived, prevents cross-user abuse)
3. Silently evicts the stale session — removes it from all state maps, aborts its NATS subscription, cancels any pending departure
4. Suppresses both `PARTICIPANT_LEFT` (old) and `PARTICIPANT_JOINED` (new)

From other peers' perspective, the user never left.

**Grace period** increased from 2s to 3s as a safety net for legacy clients that don't send `previous_session_id`.

### Multi-device safety
A user joining from two different browsers is unaffected. Each browser has its own session ID and doesn't know the other's — so neither can trigger eviction of the other.

## Files changed
| File | Change |
|------|--------|
| `videocall-client/.../connection_manager.rs` | Append `previous_session_id` to URLs on reconnection |
| `actix-api/src/constants.rs` | Grace period 2s→3s |
| `actix-api/src/lobby.rs` | Extract `previous_session_id` from WS query params |
| `actix-api/src/webtransport/mod.rs` | Extract `previous_session_id` from WT query params |
| `actix-api/src/actors/session_logic.rs` | Store and forward `previous_session_id` |
| `actix-api/src/actors/transports/ws_chat_session.rs` | Thread parameter through |
| `actix-api/src/actors/transports/wt_chat_session.rs` | Thread parameter through |
| `actix-api/src/messages/server/mod.rs` | Add `previous_session_id: Option<u64>` to `JoinRoom` |
| `actix-api/src/actors/chat_server.rs` | Eviction logic: look up previous session, verify ownership, evict silently |

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [ ] Documentation updated


darioalessandro:
## Proposed Alternative: Client-Generated `instance_id` + NATS KV

  Thanks for the thorough root cause analysis — the problem is real and well-documented. However, **the `previous_session_id` approach has a fundamental scaling flaw that we need to address
  before merging.**

  The eviction logic does local `HashMap` lookups (`room_members`, `sessions`, `active_subs`). If the client reconnects to a different server replica, the previous session won't be found and
  the eviction silently does nothing. Our Helm config already has autoscaling scaffolding — **this will break the moment we go to 2+ replicas.** We should fix this correctly now rather than
  ship something we know won't scale.

  ### Core Idea

  1. **Client generates a stable `instance_id`** (UUID) once per meeting join — survives reconnects, dies on tab close
  2. **NATS KV** (`session_instances` bucket) is the global registry: key = `{room}.{user_id}.{instance_id}` → value = `{ server_id, session_id, status }`
  3. On `JoinRoom`, server queries KV to detect reconnection — works across any number of replicas
  4. Infrastructure prereq: enable JetStream in NATS Helm values (already configured, just `enabled: false` → `true`)

  ### Key Flows

  **Same-server reconnect** (common case):
  Client reconnects with instance_id=XYZ → Server checks local cache → hit!
  → Evict old session silently, suppress PARTICIPANT_JOINED, update KV
  → Peers see nothing. 

  **Cross-server reconnect** (N replicas):
  Client reconnects to Server B with instance_id=XYZ → local cache miss
  → KV lookup: "Server A owns XYZ" → Server B claims KV, fire-and-forget
    evict notification to Server A → Server A cleans up silently
  → Peers see nothing. 

  **Multi-device** (two tabs, same user):
  Tab 1: instance_id=AAA, Tab 2: instance_id=BBB
  → Different KV keys, both coexist. No eviction. 

  **Old client / no instance_id:**
  instance_id=None → KV not queried → existing grace period logic. 

  **Server crash:**
  KV entry TTL = 60s → auto-expires. Evict notification lost, but harmless. 

  ### Why Not `previous_session_id`

  | Concern | `previous_session_id` (this PR) | `instance_id` + NATS KV |
  |---|---|---|
  | **Horizontal scaling** | ❌ Local HashMap — **breaks at 2+ replicas** | ✅ NATS KV — works at any replica count |
  | **Identity ownership** | Server generates, client caches, round-trips back | Client generates, server verifies via JWT `user_id` |
  | **Multi-device** | Implicit (absence of param) | Explicit (different UUIDs = different KV keys) |
  | **Reconnect chaining** | A→B→C, each hop must cache previous | Same UUID every time, no chaining |
  | **Stored on SessionLogic** | Permanently, but only used once | Not stored — pass-through to `JoinRoom` |
  | **Parameter name coupling** | Hardcoded in 3 places across 3 crates | Single query param, parsed at transport layer |

  ### Implementation Scope

  | Area | Changes |
  |---|---|
  | **Infrastructure** | Enable JetStream in NATS Helm values (one-line change per region) |
  | **Client** | Generate UUID in `use_hook`, append `&instance_id=` to lobby URLs, capture in reconnect closure |
  | **Server parsing** | Add `instance_id: Option<String>` to `LobbyTokenQuery`, parse from WT query params |
  | **Server threading** | Thread through `SessionLogic` → `JoinRoom` and `Disconnect` messages |
  | **Server logic** | Add KV store + local `instance_index` cache to `ChatServer`. Query KV in `JoinRoom` handler. Subscribe to `evict.{server_id}` for cross-server cleanup |
  | **Tests** | Rewrite 5 existing tests for `instance_id`, add cross-server test. Replace `sleep(200ms)` with `wait_for_condition` polling. Extract `DummySession` once at module level |

testradav:
> If the client reconnects to a different server replica,

absolutely.  good catch.  working on the solution to address replicas.

testradav:
close in lieu of https://github.com/security-union/videocall-rs/pull/854

PR #816, opened 2026-04-03
Fix host and attendants re-renders on audio level changes + refactor webtransport writes

darioalessandro:
## Audio halo causing full meeting UI re-render

The audio halo (glow effect when speaking) was causing the entire AttendantsComponent to re-render on every audio level callback. Since AttendantsComponent is the main meeting view (~1600 lines), every mic sample triggered a full grid layout recomputation, peer list iteration, and prop diffing for all PeerTile components.

**Root cause:** `audio_level: local_audio_level()` in the Host props read the signal in the parent scope, subscribing AttendantsComponent to every audio level update.

**Fix:** Audio level is now passed to Host via a Dioxus context signal (`LocalAudioLevelCtx`). A `use_effect` in Host watches the signal and writes the glow CSS directly to the DOM element via a cached `onmounted` reference. Zero component re-renders for audio level changes — just a direct `el.set_attribute("style", ...)`.

Additionally:
- PeerTile keys changed from `"tile-{index}-{peer_id}"` to `"tile-{peer_id}"`. The index caused Dioxus to destroy and recreate peer tiles when participants joined or left (shifting indices = changed keys). With stable keys, existing tiles are patched in place, preserving canvas elements, decoder state, and signal state.
- Moved `client.set_audio_enabled` / `set_video_enabled` inside their change guards so they only run when the value actually changes.

## WebTransport write methods refactor

Simplified `send_datagram` and `send_unidirectional_stream` by flattening the nested async blocks into straightforward linear flows with early returns. Same behavior, less indirection. Also ensures `writer.release_lock()` is always called before closing the transport on error.

## Other
- Renamed crate from `videocall-ui-dioxus` to `videocall-ui`
- Added dioxus-ui version to the diagnostics panel
- Removed stale `videocall-ui-dioxus` entry from `.release-plz.toml`

## Type of Change
- [x] Bug fix
- [x] Enhancement

## Testing
- [x] No tests needed

Rendering optimization — visual behavior is unchanged. Verified by confirming the "Host render" log no longer fires during speech.

### DIFF
--- a/dioxus-ui/src/components/host.rs
+++ b/dioxus-ui/src/components/host.rs
@@ -16,23 +16,23 @@
  * conditions.
  */
 
+use crate::components::{
+    canvas_generator::speak_style, device_settings_modal::DeviceSettingsModal,
+};
 use crate::constants::*;
+use crate::context::{
+    load_display_name_from_storage, save_display_name_to_storage, validate_display_name,
+    LocalAudioLevelCtx, VideoCallClientCtx,
+};
 use crate::types::DeviceInfo;
 use dioxus::prelude::*;
+use dioxus::web::WebEventExt;
 use futures::channel::mpsc;
 use gloo_timers::callback::Timeout;
 use videocall_client::Callback as VcCallback;
 use videocall_client::{create_microphone_encoder, MicrophoneEncoderTrait};
 use videocall_client::{CameraEncoder, MediaDeviceList, ScreenEncoder, ScreenShareEvent};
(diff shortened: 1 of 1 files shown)


testradav:
@darioalessandro fyi @palina-piliushyna has been working on the halo tweaks and fixes (bleed through) 

https://github.com/orgs/security-union/projects/8/views/7?sliceBy%5Bvalue%5D=palina-piliushyna

PR #839, opened 2026-04-08
R&D: all our internally consumed PRs as of Apr 7, 2026 9:27p EST

### LINKED ISSUE #572: hardware metrics
Define the minimum acceptable performance metrics for the video call project. These metrics should be based on a specified hardware configuration and include:
- The supported maximum number of participants 
- The required number of concurrent live video streams with a given video resolution 
- The required number of concurrent audio streams

### LINKED ISSUE #801: Meeting metrics to be updated in realtime
when in a meeting,
the meeting settings page for that meeting, which lists metrics that can change throughout the meeting, needs to be updated in real time.

example, 
number of participants should be auto updated when attendants join/exit the meeting without having to reload the settings page

Image

### LINKED ISSUE #805: Perf: video adaptation should start from middle ground quality rather than low end

### LINKED ISSUE #808: Perf:  content sharing needs to start from middle ground quality

### LINKED ISSUE #809: perf:  mic unmute lengthy noticed in some peers
running latest fixes with performance we noticed a few peers entry/exit throughout the meeting in meeting of 8 participants

also, peers who are geo located farthest from the server seem to be experiencing delayed mic mute/unmute

investigate what caused the random rejoins and lengthy (5 second) unmute action

### LINKED ISSUE #811: perf:  entry/exit messages with some peers

### LINKED ISSUE #814: WebTransport: per-packet uni streams cause continuous keyframe request storm

### LINKED ISSUE #818: Ability to manually select websocket vs webtransport
Image

testradav:
#829 #572
#811 
#814 
#805 
#808 
#809 
#818 
#801 

image

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [x] Documentation updated


testradav:
@darioalessandro another chunk of updates for R&D

Issue #832, opened 2026-04-07
PLI keyframe request storm on WebSocket causes bandwidth death spiral


jboyd01:
## Summary

In a 3-person WebSocket meeting on app.videocall.fnxlabs.com, continuous PLI (Picture Loss Indication) keyframe requests flood the session from the moment video streams are established, and **never recover**. This causes a bandwidth death spiral where video quality collapses to emergency tier within seconds.

This is distinct from #814 (WebTransport per-packet uni stream ordering issue) — this occurs on **WebSocket only**.

## Observed Behavior

From browser console logs (Jay's client, Edge 146 on Windows, meeting room `infra` with 2 other participants: mark.benge and christopher.heltzel):

1. Video starts normally at 1280x720, bitrate 250kbps
2. Within seconds, bidirectional PLI requests begin between all 3 participants for both VIDEO and SCREEN streams
3. PLI requests fire every ~30-35 frames and never stop
4. Adaptive quality manager detects fps_ratio=0.00 (zero usable frames) and steps down aggressively:
   - `medium` audio tier (fps_ratio=0.00)
   - `low` audio tier (fps_ratio=0.09)
   - `emergency` audio tier (16kbps, fec=true)
5. Video bitrate collapses: 250K → 118K → 91K → 71K → 55K
6. The PLI storm continues for the entire session duration (hundreds of PLI log lines)

## Key Log Evidence

```
INFO: videocall_client::diagnostics::adaptive_quality_manager — AdaptiveQuality: audio stepped DOWN to tier 'medium' (index 1), fps_ratio=0.00
INFO: videocall_client::diagnostics::adaptive_quality_manager — AdaptiveQuality: audio stepped DOWN to tier 'low' (index 2), fps_ratio=0.09
INFO: videocall_client::diagnostics::adaptive_quality_manager — AdaptiveQuality: audio stepped DOWN to tier 'emergency' (index 3), fps_ratio=0.00
```

PLI pattern (repeats hundreds of times):
```
INFO: videocall_client::decode::peer_decode_manager — Sending KEYFRAME_REQUEST to mark.benge@hcl-software.com for VIDEO
INFO: videocall_client::client::video_call_client — Received KEYFRAME_REQUEST from christopher.heltzel@hcl-software.com for VIDEO
INFO: videocall_client::encode::camera_encoder — CameraEncoder: forcing keyframe at frame NNNN (PLI)
```

## Root Cause Hypothesis

[...]

PR #826, opened 2026-04-03
Feature:  select protocol

testradav:
Ability to select webtransport or websocket manually
We needed this feature to debug errors while using the different protocols

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [x] Tests written or updated
- [ ] No tests needed

## Other
- [ ] Documentation updated
- [ ] Before/After Images provided

image

image

image

PR #854, opened 2026-04-14
RnD sync — April 3 to April 27 

### LINKED ISSUE #452: Renaming your user on the client causes you to leave/rejoin the meeting

### LINKED ISSUE #742: Option to allow admit message by anyone in the meeting vs just the host

testradav:
# April 3

- these are the PRs we have taken internally from upstream that have yet to be merged 
- these have been tested internally and are currently deployed in our internal servers
- this R&D PR should be merged into main; going the R&D route as discussed. The PRs linked below can be closed or merged (will wash out as the code is already in this PR)

- **feat: allow admit messages from any participant or host only (#765)** — `admitted_can_admit` toggle so admitted participants can gate-keep the waiting room on behalf of the host, closes #742.
- **feat: update share screen functionality (#766)** — improvements to the screen-share flow by @Alena-Melnikava.
- **fix: renaming user on the client causes you to rejoin the meeting (#776)** — fixes the regression where renaming your display name forced a full rejoin; closes #452.
- **feat: select protocol (#826)** — UI affordance to manually select between WebSocket and WebTransport for a session.

# April 7

### initial enterprise-driven performance + observability batch

- **feat: vcprobe diagnostics, health reporter, relay Prometheus metrics (#829)** — new diagnostics probe + client-side health reporter + server-side relay Prometheus metrics for end-to-end quality observability.
- **feat: hardware metrics (#572)** — per-peer hardware-usage metrics pipeline.
- **perf: entry/exit dropped messages with some peers (#811)** — fixes participants seeing stuck entry/exit messages for specific peers.
- **fix: WebTransport per-packet uni streams cause continuous keyframe-request storm (#814)** — breaks the feedback loop where per-packet uni streams under WebTransport triggered continuous keyframe requests.
- **perf: video adaptation starts from middle-ground quality (#805)** — AQ now starts at a middle tier rather than the low end, delivering immediately-acceptable quality before climbing.
- **perf: content sharing starts from middle-ground quality (#808)** — same pattern for screen share — start at middle tier rather than low.
- **perf: long mic-unmute delay for some peers (#809)** — fixes the long latency some peers saw when unmuting their mic.
- **feat: ability to manually select WebSocket vs WebTransport (#818)** — user-level transport preference (precursor to the UI work in #826).
- **feat: meeting metrics updated in realtime (#801)** — per-peer metrics update live in the UI rather than batch.

# April 8

### cleanup; performance adaptation fixes and improvements 

- **helm: enable ingress by default for meeting-api and websocket** — Kubernetes ingress is now on-by-default on both deployments so PR-preview slots and local clusters no longer need manual port-forwarding or per-env overrides.
- **Remove Matomo analytics integration (#441)** — removes the Matomo tracking code and dependencies; the integration wasn't being used and added weight to the WASM bundle.
- **helm: refactor grafana into shared base chart with per-env values** — extracts the Grafana config into a reusable base chart so per-environment values files layer on top, eliminating copy-paste drift between us-east / sandbox / preview.
- **fix: clamp audio packet loss metric to prevent impossible values** — clamps `audio_packet_loss_pct` to `[0.0, 1.0]` at the emitter so transient overflow during cadence glitches doesn't poison Grafana panels and alerts.

# April 10

### performance adaptation fixes and improvements, UX screen sharing, per peer metrics 

- **feat: receiver-side metrics for quality observability (#572)** — per-peer receiver-side metrics (received FPS, lost / out-of-order packets, jitter aggregates) feeding the AQ loop and observability dashboards so quality incidents can be diagnosed from the receiver's side, not just the sender's.
- **fix: break PLI keyframe-request storm death spiral (#832)** — breaks the loop where packet loss triggered PLI → keyframe burst → more loss → more PLI → spiral. Adds PLI rate limiting on the encoder side to bound bursts.
- **fix: camera + screen-share bandwidth coordination** — when screen share starts, the camera AQ ceiling drops so both streams share the uplink without thrashing; camera ceiling restored when screen share ends.
- **fix: display-name collision when same account joins from multiple devices** — disambiguates the display name shown to peers when the same OAuth account is connected from multiple browsers / devices so participants don't see two indistinguishable tiles with the same label.
- **fix: screen-share crash (wrong tier index in AdaptiveQualityManager)** — AQ crash fix; the screen-share tier index was being computed against the camera tier table, producing an out-of-bounds read when table lengths differed. Uses the correct screen-tier lookup now.
- **fix: WT close race errors and RefCell borrow conflict on election** — fixes a WebTransport close race (session `close()` invoked twice under tight timing) and a `RefCell` borrow conflict triggered when connection cleanup runs concurrently with a candidate swap during re-election.

# April 14

### performance adaptation fixes and improvements, UX tiling, speaker highlighting  

- **fix: User can hear other users when he's in in waiting room (#736)** — privacy fix. Stops audio from active-meeting participants reaching a user who is still in the waiting room. Client-side filter plus a server-side allowlist (only `MEETING` and `SESSION_ASSIGNED` PacketTypes forwarded to observers; everything else fails closed).
- **feat: tier transition tracking, screen video metrics, quality debugging dashboard** — Prometheus metrics for AQ tier transitions plus separate screen-video stats, plus a Grafana debugging dashboard for diagnosing quality incidents.
- **fix: Speaking glow bleed-through** — fixes the visual issue where glowing borders from other participants were visible through a pinned screen share.
- **fix: Tile distribution for large meetings (#740)** — density-mode support (`Auto` / `Dense` / `Maximum` / `Standard`) plus an overflow-count badge so large meetings keep a usable tile layout instead of shrinking tiles past readability.

# April 16

### lots of performance adaptation fixes and improvements well as UX tiling, speaker highlighting 

- **fix: tile distribution for large meetings (feature stub #740)** — (continues from April 14) further polish on the density-mode overflow-count layout work.
- **feat: automated console log collection and upload** — opt-in `consoleLogUploadEnabled` flag that ships buffered browser console logs to a server endpoint for incident triage, with PII scrubbing, byte-budget flushing, and heap / platform preamble.
- **fix: replace worst-peer FPS with p75 percentile aggregation** — the AQ FPS input is now a p75 aggregate across peers rather than the minimum, smoothing the signal and avoiding tier thrash when a single briefly-stalled peer drags the aggregate.
- **fix: re-election fallback — abort when winner RTT is not better** — if the re-election winner's RTT isn't actually better than the current connection's, abort and restore the prior measurement instead of swapping into a worse transport.
- **feat: persistent connection-quality warning for high RTT** — sustained high-RTT state now surfaces a persistent visual warning (Good / Warn / Critical) instead of flashing between states per-sample.
- **fix: PR-G observability cleanup — complete audio metric rename** — finishes the audio-metric renaming started earlier so Prometheus / Grafana label names are consistent across the fleet.
- **feat: bot WebSocket support, conversation mode, and RX diagnostics** — bot can now join over WebSocket (not just WebTransport) in a multi-participant conversation mode, and logs per-packet RX diagnostics for scripted test scenarios.
- **fix: handle user display name renaming (#243)** — surfaces display-name changes in the UI after a rename; live-updates across participants via `PARTICIPANT_DISPLAY_NAME_CHANGED` events.
- **feat: screen-share warmup (8s) and 720p quality floor** — initial 8-second screen-share warmup at 720p so text is readable from the first frame; floor stays at 720p even at minimum quality.
- **fix: add missing `on_toggle_pin` prop to mock PeerTile** — build fix after the pin feature landed; mock-tile construction was missing the new required prop.
- **feat: climb-rate limiter to prevent adaptive quality yo-yo (PR-H)** — ceiling / dwell / backoff machinery that prevents the AQ climber from immediately stepping back up after a crash-triggered step-down. Breaks the yo-yo loop where repeated bandwidth drops cause visible quality thrash.
- **fix: lower video recovery threshold for parity with audio (PR-D)** — lowers the video recovery FPS threshold from 0.85 to 0.70 so video recovers at the same hysteresis boundary audio already uses; Grafana dashboard + docs updated to match.

# April 22

### guest meetings, host-leave lifecycle, WebSocket AQ, join-page redesign, selective decode, CONGESTION routing, console-log hardening

- **feat: introduce support of meeting with guests** — guests join via a `guest:{uuid}` identity, bypass OAuth, and go through the waiting room with a `is_guest` JWT claim.
- **feat: "End meeting when host leaves" toggle** — when the host disconnects and the toggle is on, the meeting broadcasts `MEETING_ENDED` cleanly with `PARTICIPANT_LEFT` ordering, plus a TOCTOU-safe host-gone check folded into the `join_attendee` transaction so late joiners aren't stranded.
- **fix: PID integrator stuck at maximum with no recovery** — adds a 30s saturation watchdog that force-resets the integrator when `pid_output` sits at max too long, breaking the feedback loop where reduced bitrate caused low received FPS.
- **fix: tile layout bounce + GUID display-name detection** — deterministic tiebreaker in the speech-priority sort so tiles with equal timestamps no longer shuffle; IdPs returning a UUID as the name claim now fall back to a human-readable name derived from the email.
- **feat: enable adaptive quality for WebSocket-only users (OSS #859)** — ports the WebTransport bounded-channel + CONGESTION-via-NATS pattern to WebSocket, plus client-side self-detection of WS backpressure that force-steps-down video quality on local drops.
- **fix: console log upload collision + retry storm** — widens `uuid_short` from 8 to 16 hex chars; server returns 409 Conflict on EEXIST instead of 500; client drops retry/requeue entirely (fire once, log-and-drop on failure) to prevent storms during backend outages.
- **feat: redesign join page in same visual style as other pages** — extracts settings toggles into a new `PreJoinSettingsCard` component.
- **fix: add missing proto files supporting the guest-meeting feature.**
- **fix: match CONGESTION signals against session_id history to survive re-election** — bounded 16-entry `VecDeque` so signals targeted at a prior session_id still match, while preserving per-tab / per-device isolation.
- **fix: cap large-meeting decode load to active layout** — replaces per-canvas `IntersectionObserver` visibility with a layout-driven `active_decode_set`; off-screen peers skip video/screen decode but audio is always decoded.
- **fix: add missing `sender_user_id` arg to WebSocket `on_outbound_drop` call** — restores the build after the CONGESTION-routing signature widening.

# April 23

### server-side display-name hardening, stable tile ordering, screen-share fixes

- **security: per-user rate limiter on display-name changes** — 5/min budget shared between `PUT /display-name` and `POST /join` to close the leave-and-rejoin bypass; NATS packet sanitization for `PARTICIPANT_DISPLAY_NAME_CHANGED` events (UTF-8 fail-closed + server-side `validate_display_name` + authoritative `room_id` rewrite before forwarding); static error messages to prevent reflected-input info disclosure; client-side revalidation of server-sent display-name changes.
- **fix: stable join-time tile ordering with overflow-speaker promotion** — replaces speech-priority sort (which shuffled the grid as people took turns speaking) with stable join-time ordering; loud speakers in the overflow set are swapped into the visible set by displacing least-recently-active visible peers; auto-density escalation bumps density mode automatically so every active speaker stays on-screen.
- **fix: activate shared screen for newly-joining participants** — when a new participant joins a meeting with an active screen share, the client now fires a keyframe request (PLI) on the visibility transition so the shared content renders instead of staying blank; also tightens the WS `on_outbound_drop` path with the cleaner `if let Ok(pw)` pattern.
- **fix: improve tile grid + placement during screen share** — better layout when someone is sharing screen; `active_decode_set` branches on `has_screen_share` to use the screen-share-panel tile set vs. the grid tile set appropriately.

# April 24

### searchable meeting names, appearance personalization, screen-share quality, console-log hardening, WebTransport upgrade

- **feat: searchable meeting names with enforced access control** — meetings are indexed into a SearchV2 backend with per-participant ACLs derived from the `meeting_participants` membership set; the UI gets a new search modal that honours the caller's principal so users only see meetings they're actually part of. Adds Lucene-metachar escaping on the query path and a prefix+multi-match hybrid so "MyMeeting" still matches case-insensitively while the stored identifier stays case-preserving.
- **feat: participant-chosen speaker-highlight glow** — new "Appearance" tab in Settings lets a viewer pick their own speaker-highlight color (preset or custom hex), tune outer/inner intensity, or turn the effect off entirely. Preferences persist in localStorage with a 300ms debounce and a drop-guard that cancels pending writes if the user navigates away mid-change. The runtime `speak_style` and the settings-panel preview share a single `calculate_glow_params` helper so the colors match exactly.
- **feat: resizable screen-share / participants split with pinning parity** — screen-share layout now exposes a draggable vertical handle (clamped to 30%-85%) between the shared content and the participants panel; every tile (including the screen-share tile itself) gets a pin affordance; stopping the share restores the normal grid without leaving layout artifacts behind.
- **feat: default screen-share quality raised to 1080p with VBR** — screen-share encoder now starts at 1080p / 2500 kbps ideal (4000 kbps VBR peak), sets `contentHint = 'detail'` on the MediaStreamTrack so WebCodecs optimizes for text sharpness rather than motion, and wraps `getDisplayMedia` constraints in `{ideal: N}` dictionaries so non-1080p monitors (1440p / 4K / ultrawide) aren't rejected. Adaptive-quality step-down still owns the correction path on constrained uplinks.
- **fix: upgrade web-transport-quinn 0.8.1 → 0.11.9, fix WT inbound reception on bot and vcprobe** — closes a `quinn`-side `accept_uni` waker bug that silently dropped inbound streams, and replaces the old "one stream = one packet" `read_to_end` pattern in the bot / vcprobe WT clients with an explicit length-prefixed frame loop that matches the relay's persistent-stream wire format. Validated on a 6-bot 50/50 WT/WS load test.
- **fix: console log upload restored after CORS regression** — adds `x-chunk-seq` to the server's CORS `allow_headers` list so browsers accept the POST after preflight. Zero console logs were being collected for ~24h before this landed.
- **feat: gzip console log chunks at write time** — server compresses each uploaded NDJSON chunk with `flate2::Compression::fast` on a `spawn_blocking` task, stores as `.log.gz` instead of `.log`. ~10:1 ratio on text logs, ~10× less disk I/O per chunk, 5 GB PVC retention stretches further. Quota stays measured in uncompressed bytes so the limit can't be gamed with highly-compressible input; retention cronjob now matches both legacy `.log` and new `.log.gz` files so the transition is seamless.
- **feat: costume-video bot + load-testing infrastructure** — swap the bot's trivially-compressible EKG-waveform video for pre-recorded Google Meet costume clips driven by audio RMS (idle vs. talking) so VP9 encoding and server relay bandwidth are measured against realistic compression. Observer-mode bots connect and send health packets without media producers, for capacity testing without wasted CPU on silent audio.
- **feat: DEV_USER auto-login for local development** — new `DEV_USER` env var (`email:display_name`) enables a dev-only `GET /api/v1/dev/auto-login` endpoint that issues a session cookie when OAuth is disabled and redirects to `/`. Triple-gated (config-time OAuth-disabled check, request-time `dev_user.is_some()` check, startup warn log) and returns 404 in any production-style configuration — invisible to external probes.

# April 27

### resizable screen-share split, dock-style controls bar, live-sync admitted_can_admit, helm prom-annotation cleanup

- **feat: resizable screen-share area + pin-on-share for every tile** — adds a draggable vertical handle (clamped to 30%-85%) between shared content and the participants panel; every tile (including the screen-share tile itself) now has a pin affordance; right-panel column count collapses to 1 when the right panel is ≤25% of viewport or physically narrow; stopping the share restores the normal grid without leaving layout artifacts. Sibling unified `ss_tile_h` formula keeps tile height fixed at 4-rows-per-column regardless of resize position so tiles don't bounce as the user drags.
- **feat: Mac-dock-style auto-hiding meeting-controls bar** — primary buttons (mic / cam / hangup) stay rendered always; secondary buttons (screen-share, peer-list, diagnostics, settings) collapse after 1s of inactivity; the whole bar fades to 10% opacity after 4s and reveals on hover (mouse) or tap (touch). New dock-position dropdown lets the user dock the bar at the bottom, left, or right of the viewport. Touch handling via `touchstart` listener so mobile users can tap-to-reveal.
- **fix: live-sync `admitted_can_admit` mid-meeting via NATS settings event** — closes a real bug where `admitted_can_admit` was a static prop snapshot at join-time, so toggling it mid-meeting had no effect on already-joined participants. Adds a new `MEETING_SETTINGS_UPDATED = 10` protobuf event published by `meeting-api` on every meeting PATCH, a client callback (`on_meeting_settings_updated: Option<Callback<()>>`) that triggers a status refetch, and a reactive signal in the UI that updates host-control visibility live. Also fixes guests not always receiving an observer token (needed for the unauthenticated status refetch) and the waiting-room callback over-triggering settings refresh.
- **feat: read-only observer token for guest status refetch** — to make the live-sync refetch above work for unauthenticated guests, the server now mints a guest-side observer token on `/join-guest`. Token has `room_join: false, observer: true, is_guest: true` claims, is meeting-bound, time-bounded (30-min TTL), and the server enforces a cross-meeting check on every refetch (rejects with 401 if the URL `meeting_id` doesn't match the token's claim). Media-server packet allowlist drops everything except `MEETING` and `SESSION_ASSIGNED` for observer sessions — pure read access.
- **fix(helm): remove hardcoded prometheus scrape annotations from meeting-api** — the `meeting-api` Axum service does not expose a `/metrics` endpoint (no prometheus crate dep, no `/metrics` route in the router). The hardcoded `prometheus.io/scrape=true` annotation in `helm/meeting-api/templates/deployment.yaml` was causing Prometheus to scrape `:8081/metrics` and receive HTTP 404, firing the "scrape target down" alert and skewing the up{} count for the videocall namespace. The `{{- with .Values.podAnnotations }}` override block is preserved so callers can still inject annotations via values once a metrics surface is added.
- **chore: drive-by clippy cleanups in `attendants.rs` (`promote_speakers`)** — `&mut Vec<String>` → `&mut [String]`, range-loop → `iter().enumerate().skip()`, `.map_or(true, …)` → `.is_none_or(…)`, multi-line sort closures collapsed into single-line forms. Closes 4 of 6 outstanding clippy warnings.

# April 27 — late-day batch

### participant-frame highlight bug fixes, home-page UX cleanup, idle-meeting auto-activation regression tests

- **fix: participant frame border resets to default after speaking stops** — closes a regression in the speaker-highlight glow where the inline `border-color` from a prior speaking event got "stuck" on the tile. Single `DEFAULT_TILE_BORDER_COLOR` const now drives both the runtime `speak_style()` reset path and the appearance-settings preview, so they're guaranteed identical. Also fixes the host's participant tile not getting the speaking glow while screen-sharing — the screen-share *content* tile correctly stays glow-free, but the host's face tile is now eligible for highlight (removed an over-broad suppression condition). Glow toggle visuals updated to Apple iOS green (ON) / gray (OFF).
- **feat: Playwright coverage for getDisplayMedia and fake-mic** — `speaker-highlight.spec.ts` previously had two "cannot automate" gaps. This PR closes both: (a) a synthetic `getDisplayMedia()` shim via `addInitScript` that resolves to a `canvas.captureStream(12)` driven render loop, bypassing the OS picker; (b) `--use-file-for-fake-audio-capture` plus a real WAV fixture so remote speaking can actually be observed, not just inferred. Tests now exercise both the speaker-highlight reset-after-mute path and the host-stays-highlighted-while-sharing path end-to-end.
- **feat: home-page create/start meeting screen revamp** — replaces the conflicting "Create a New Meeting" + "Start or Join Meeting" pair with a mutually-exclusive flow keyed off the meeting-id input. Empty field shows only `Generate a New Meeting ID`; filled field shows only `Start or Join Meeting`. Generate now populates the meeting-id field (and saves display name to localStorage) rather than navigating directly into the meeting, so the user explicitly clicks Start/Join to enter. Static "allowed characters" help-text under both inputs is replaced with per-keystroke inline validation that surfaces only when an invalid character is typed (e.g. `'@' not allowed — only letters, numbers, …`) and reuses the canonical `is_allowed_display_name_char` predicate so client- and server-side validation stay synchronized.
- **test: integration coverage for non-host idle/ended-meeting auto-activation authorization** — adds 5 integration tests in `meeting-api/tests/wr_idle_ended_join_tests.rs` covering the four legitimate scenarios (idle + WR-off → activates and admits; ended + WR-off → re-activates and admits; idle + WR-on → `waiting_for_meeting`; WR-toggle race during join → `admitted` or `waiting`, never the broken stuck state) plus an explicit security-regression guard (`test_non_host_cannot_activate_meeting_with_wr_on`) that verifies a non-host cannot bypass the WR gate by triggering activation. Locks in the authorization invariants previously documented only in code comments.

---

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Enhancement
- [x] Breaking change
- [x] Project / Infra

## Testing
- [ ] No tests needed

## Other
- [x] Documentation updated


testradav:
@darioalessandro  our R&D branch had two problems

1. it had been polluted with hcl internal code; some of it was redoing the git workflows since the internal pipelines are different from public
2. the & within the branch name was creating all sorts of problems when using github online to view/edit files

I have since 
- recreated our `R&D` branch clean as `RnD` off of `main`
- recreated the PR that has all our R&D code that we have tested and are running with internally

testradav:
:arrow_right: **April 24 sync pushed** — added a new section summarizing what landed on PR-staging this cycle: searchable meeting names, participant-chosen speaker-highlight glow, resizable screen-share layout + 1080p default quality, `web-transport-quinn` 0.11.9 upgrade (closes the `accept_uni` waker bug that was dropping inbound WT streams), gzip-at-write for console logs, CORS fix for console-log uploads, costume-video bot load-test infrastructure, and DEV_USER auto-login for local development.

Read the new `# April 24` section in the body for the full bullet list.

testradav:
:arrow_right: **April 27 sync pushed** — added a new section covering: resizable screen-share split with per-tile pin (incl. the share-tile itself), Mac-dock-style auto-hiding meeting-controls bar with touch tap-to-reveal, live-sync of `admitted_can_admit` via a new `MEETING_SETTINGS_UPDATED` protobuf event + read-only observer-token issuance for guest status refetch, helm meeting-api prometheus-annotation cleanup, plus drive-by clippy cleanups in `attendants.rs` that closed 4 of 6 outstanding warnings.

Read the new `# April 27` section in the body for the full bullet list.

testradav:
:arrow_right: **Late-day batch on April 27** — second sync today. New section in the body covers: speaker-highlight border-reset fix + host-glow-while-sharing fix + Playwright coverage closing the longstanding `getDisplayMedia` and fake-mic automation gaps; home-page create/start screen revamp (mutually-exclusive Generate/Start-or-Join with inline validation); 5-test integration suite for non-host idle/ended-meeting auto-activation authorization (incl. an explicit security-regression guard).

Read the new `# April 27 — late-day batch` section in the body for the full bullet list.

PR #862, opened 2026-06-04
RnD sync — May 5 to June 7

testradav:
# June 7

### RnD sync — load-test bot now exercises simulcast

- **The synthetic load-test bot publishes simulcast layers.** The "costume" / EKG load-test bot now publishes up to 3 simulcast video layers (L0/L1/L2) like the real browser client, instead of only ever sending the base layer. Bot-driven load tests now exercise the relay's per-receiver layer selection and forwarding path end to end — the same path the June 5–6 client/relay simulcast work introduced.

# June 6

### RnD sync — simulcast sender-side decoupling, theming, low-end UX

Completes the per-receiver simulcast work from the June 5 cycle by removing the last path through which one struggling participant could degrade quality for the whole room, and adds light/dark theming plus low-end-device and logging improvements.

#### Simulcast & adaptive quality
- **The sender no longer adapts to receiver-reported frame-rate.** Previously the publisher shed simulcast layers based on *aggregated receiver* FPS, so a single degraded receiver could collapse a quality layer for everyone in the room. The publisher now reacts only to its own local signals — encoder-queue backpressure and server-reported congestion — so a struggling receiver simply pulls a lower layer for itself and can no longer drag a layer down for the rest of the room. This is the core decoupling the per-receiver simulcast design was building toward.
- **Publish-side layer suppression.** The relay now computes the highest layer any receiver actually wants for each publisher and hints the publisher to stop encoding layers nobody is watching — saving encode CPU and uplink bandwidth — then restores them eagerly the moment a receiver asks for more. Fail-open: a missing or stale hint never starves a receiver.
- **Live simulcast / adaptive-quality diagnostics.** A live, per-section view in the Performance settings panel showing in real time what per-receiver simulcast and adaptive quality are doing for each of send and receive, across video, audio, and screen-share. Additive observability only — no behaviour change.
- **Layer-transition logging.** One log line per change in the local sender's active layer count, reporting per-layer resolution, applied bitrate, and ACTIVE/SHED state.

#### Theming
- **Light and dark themes.**

#### Performance & low-end devices
- **Removed the pre-join CPU-capability hard block.** Low-end machines (2-core / older Intel Macs) were blocked from joining with a "Device not supported" message despite working fine in practice — the block is gone.
- **Per-packet console logs demoted to `trace`,** plus a runtime `logLevel` dial so client log verbosity can be raised or lowered without a code change.

# June 5

### RnD sync — per-receiver simulcast + observability

- **Per-receiver simulcast (adaptive quality / multicast).** Each participant now publishes several independent quality layers (video, audio, and screen-share), and every receiver pulls the layer its *own* network and device can sustain — so one peer's poor connection no longer drags the whole room down to a lowest common denominator. Layer selection is receiver-driven (per-peer downlink estimation with fast-down / conservative-up hysteresis and a sequence re-anchor on switch to avoid self-induced loss), and the SFU relay forwards the chosen layer per receiver. A new in-meeting **Performance** settings panel lets each user set min/max quality **thresholds** for both what they **send** (cap upload + encode CPU) and what they **receive** (cap download), with live "Sending" / "Receiving" needles and a Send | Receive toggle. Multicast is enabled by default and gated by a per-device capability ceiling, so low-power devices automatically fall back to a single layer.
- **Relay observability metrics.** New relay-side meeting-health metrics — slow-receiver detection, inbound-mailbox pressure, and re-election signals — for operator triage, plus meeting-investigation dashboard repairs.
- **Bot WebTransport framing fix.** The load-test bot now length-prefixes its WebTransport upstream packets (`[u32 BE len][payload]`) to match the relay's framed reader, fixing a silent upstream blackhole.
- **E2E.** Added Performance-panel end-to-end coverage; the multi-party simulcast e2e is quarantined pending a renderer-resilient CI runner (capability-override + multi-context stability).

# June 4

### RnD sync — internally-deployed work since the May 5 cycle

Batches all HCL-internal rollups that landed on the integration branch since the prior sync. Highlights by area:

#### Transport & connection reliability (WebTransport / WebSocket)
- Major re-election + connection-preservation hardening: preserve the existing connection on total candidate failure, retry re-election when candidates were transiently unavailable, treat sustained implausible RTT as a re-election trigger, suppress re-election on a CPU-stalled main thread (drift watchdog), and recover from the "only 1 server configured" stranding case.
- Distinguish handshake failures from mid-session drops in connection-loss reporting.
- Per-media-type persistent WebTransport streams with split writer tasks; outbound-channel sizing tuned (default capacity raised) with a priority-aware drop policy and congestion-aware keyframe exemption; relay actor mailbox capacity aligned to the outbound channel to prevent mailbox-full drops.
- Move the heartbeat off the main thread so it survives background-tab throttling.
- Encoder auto-recovery on closed-codec / VPX memory errors; drop stale buffered video frames and skip to live.
- Protocol settings simplified — removed the "Auto" option now that WebTransport includes an automatic WebSocket fallback.

#### Adaptive quality & simulcast (foundation)
- Per-receiver simulcast groundwork: additive `simulcast_layer_id` on the packet wrapper, the client N-layer publisher scaffold + receiver layer guard, and per-receiver layer selection in the relay (flag-gated/dark).
- Viewport-aware relay filtering so subscribers only receive tiles they render.
- AQ refinements: PID slew-rate limiting + tighter encoder-reconfigure threshold, aggressive bitrate cut/hold on self-targeted congestion, an outlier guard so one constrained receiver can't poison all senders, source-aspect-ratio preservation when fitting to quality tiers, and an adaptive decode-budget for weak hardware.
- Freeze-blind observability: surface per-peer loss + keyframe rate into the quality score, and decode-budget state into the health packet.

#### Host controls
- Immediate, host-excluded propagation of mute-all / disable-video-all to every participant (cut from ~5s to sub-second).
- Disable video for all or a specific participant; mute all or individual participants; kick a participant.
- Toggle for entry/exit toast messages and sounds.

#### Meetings list, home & pre-join
- Meetings-list filter + sort (ownership / status / attended-within; sort by last-active, last-attended, owned, etc.) with persistence.
- Pre-join device preview: live camera preview, camera/mic toggles, mic level meter, and camera/mic/speaker selectors that carry into the meeting.
- Meeting lifecycle: presence-driven `idle` state when empty, re-activate on join.
- About modal + version footer on the homepage; host identity in the meetings-list tooltip.

#### Peer-tile & layout
- Consistent peer-tile aspect handling (cap at 3:2), clip the video feed to the tile's rounded corners, preserve source aspect ratio, and refine 2-peer fullscreen / grid sizing.
- Dock-position selector + control-hiding toggle, persisted to local storage.

#### Screen-share
- Screen-share quality metrics in the peer diagnostics panel and downscale-cause surfacing; a visibility toast with peer-confirmed ack; consumer-side aspect-ratio preservation; hold last-known FPS/kbps during static periods.

#### Diagnostics & signal meters
- Per-peer transport (WT/WS) badge in diagnostics and signal popups; floating, source-anchored signal-meter popups with multi-open support and drag affordance.

#### Security
- Redact `?token=<JWT>` from browser-console + state-callback URL surfaces and scrub the active-server URL from the health packet before publish.
- Server-side enforcement of waiting-room audio isolation (+ adversarial tests across media types and non-allowlisted packets, WT-path parity).
- Require explicit opt-in for unverified OAuth; `ALLOW_ANONYMOUS` removed (defaults off, `DEV_USER` for local dev).

#### Observability / telemetry
- Client observability telemetry (loss/keyframe/encoder counters, Long Tasks API observer, capability benchmark); fixed a participant-gauge leak (derive from live sessions); relay Prometheus counters for auth rejections + outbound-channel drops; Grafana dashboards + alerts for client heap and datagram drops.

#### Theming & UI
- Token-based styling replacing hardcoded values, foundation for runtime theme switching, and a color picker for appearance.

#### Testing
- Comprehensive E2E repairs (0 failures), fast bvt0/bvt1 smoke projects, and broader coverage across host controls, waiting-room isolation, and device selection.


testradav:
Pushed a new **June 5** section — per-receiver simulcast (adaptive-quality multicast + Send/Receive thresholds panel), relay observability metrics, bot WebTransport framing fix, and Performance-panel e2e. Merge is clean (no conflicts); meeting-api + both wasm crates build green.

testradav:
Pushed a new **# June 6** section (through commit `6f802a1f`). This cycle completes the per-receiver simulcast work from June 5:

- The **sender no longer adapts to receiver-reported FPS** — a single struggling receiver can no longer collapse a quality layer for the whole room; the publisher reacts only to its own encoder backpressure + server congestion.
- **Publish-side layer suppression**: the relay hints the publisher to stop encoding layers no receiver wants (saves encode CPU + uplink), fail-open and restore-eager.
- **Live simulcast/AQ diagnostics** in the Performance panel, and per-layer transition logging.
- **Light & dark themes**, removal of the pre-join CPU-capability hard block, and per-packet logs demoted to `trace` with a runtime `logLevel` dial.

PR #865, opened 2026-06-17
RnD sync

testradav:
# June 18

Follow-up R&D batch on top of the June 17 cycle — decode-budget and PLI-backoff hardening, a few perf memoizations, a new battery-level metric, a localStorage-format migration, and a round of test/CI tidies. OSS CI workflows are restored from `main` and internal-only ops/deploy files remain excluded from this branch.

### Adaptive video & transport robustness

- **Pinned-peer decode admission is now bounded against the decoded budget.** A pinned (always-decode) peer could be admitted beyond the decode cap in the phase-3 admission pass; it's now clamped against the same `decoded_bucket` accounting as everyone else, so a pin can't overcommit the decoder.
- **Proactive-PLI backoff invariant hardened.** Tightens the exponential-backoff invariant on the proactive keyframe-request (PLI) path so repeated requests can't collapse the backoff window, plus a couple of stale-comment corrections.

### Audio playout & NetEQ

- Delay-manager and NetEQ correctness tidy-ups and a jitter-buffer refinement carried forward from the resync-governor work.

### Observability

- **`client_battery_level` is now exported as a numeric Prometheus gauge** for fleet battery observability (previously not surfaced as a metric).

### Storage & DX

- **Legacy `localStorage` values migrate from zlib-compressed blobs to plain JSON on load.** Older releases persisted some keys zlib-compressed, which made them opaque to debugging and hand-editing; a one-time, magic-byte-gated migration rewrites them to readable plaintext without touching already-plaintext values. Covered by a new e2e migration spec.

### Performance

- **`window.__APP_CONFIG` is parsed once and memoized** instead of being re-deserialized on every render through the `RuntimeConfig` accessors.
- **Parsed bundled theme JSON is cached across theme applies** so toggling themes no longer re-parses the theme file each time.

### UI

- Unified-timeline crosshair tooltip is now hidden when the diagnostics chart unmounts (no lingering tooltip after the drawer closes).
- Speaker-highlight swatch delete button inset so the container's `overflow: hidden` no longer clips it; dead/duplicate settings-modal CSS removed.

### CI & tooling

- Added a repo `.gitleaks.toml` and a `.pre-commit-config.yaml` (secret-scanning + pre-commit hooks).
- A clippy-CI self-consistency check (`scripts/check-clippy-ci-sync.sh`) that keeps the lint gate from silently drifting.

# June 17

Sync of the latest R&D batch from the internal integration branch. This is a large cycle — the highlights below are grouped by area; OSS CI workflows are restored from `main` and internal-only ops/deploy files are excluded from this branch.

### Adaptive video: decode budget & tile rendering

- **Camera-off peers excluded from the decode budget.** A camera-OFF remote peer produces no video to decode, but was still counted against the decode cap — rendering with a dashed "off-budget" outline and stealing a budget slot from a real camera-ON stream. Camera-on/off tiles are now partitioned before the budget split; cameraless peers render as plain avatars and never consume a slot. A peer that joins camera-off and turns the camera on mid-call now starts decoding on an unpressured host.
- **Reliable un-pause affordances for budget-shed videos.** Adds a per-tile PLAY button on "Video paused" tiles (force-decodes that peer, clamped by the device/iOS decoder ceiling), a persistent "Show all videos ↔ Back to automatic" toggle in the Appearance panel, and a true "All" decode override.
- **Per-peer transport badge (new, server-gated, default off).** Each remote peer tile can show a small "WT"/"WS" badge next to the signal meter indicating whether that peer's media is on WebTransport or WebSocket — opt-in via runtime config.
- Screen-share decode-budget cleanup and a bounded pinned-tile swap path.

### Audio playout & NetEQ

- **Audio resync governor** for the audio-ahead-of-video drift case: NetEQ had no resync-to-live governor, so the adaptive target could ratchet and accumulate multi-second playout latency. Adds an observable audio-playout-latency gauge plus operator dashboard panels and alert rules.
- NetEQ sequence-number-wrap and sample-timestamp correctness fixes, a flush path fix, and an audio-congestion downshift.

### Transport resilience, reconnect & parity

- **Suppress replayed "peer left/joined" events during a transport reconnect.** The server replays the member list on reconnect, which previously produced a storm of spurious leave/join toasts and sounds after a network blip (~30 toasts in a 15-person meeting).
- Receiver-side downlink-congestion signal, proactive PLI backoff, reconnect escalation, RTT-probe resilience, and a WebTransport/WebSocket behavior-parity audit so both transports converge on the same reconnection semantics.

### Host controls & meeting lifecycle

- **Grant / revoke host** to a participant mid-meeting (new host-granted / host-revoked lifecycle events).
- End-on-host-leave join-guard hardening and improved peer-leave decode handling.

### Observability & diagnostics

- Layer-switch ↔ freshness-skip correlation logging to attribute decode freshness skips to recent simulcast layer transitions.
- Decoder Web-Worker `log` output forwarded into the main-thread console-log upload buffer so worker logs reach field diagnostics; a diagnostics timeline view.

### UI / theming

- Theme-file support (MVP) and custom meeting colors, plus fixes for custom-color overflow into the preview and an over-aggressive background blur when the settings panel is open.

### Performance & CI

- Relay hot-path tidy-ups (gate the at-cap forced stale sweep; exact varint32 `user_id` read).
- Backend CI timeout, cold-compile/sccache, and clippy-gate improvements.


testradav:
📦 Pushed the **June 18** R&D cycle (`7cb47252`) — decode-budget/PLI-backoff hardening, perf memoizations, a battery-level metric, a localStorage-format migration, and CI/test tidies. 36 files, +1619/−485. New section is at the top of the PR description; OSS workflows restored from `main`, internal-only files excluded.

testradav:
Resolved the merge conflict with the new base (`#864` landed on `RnD`).

`#864` was an earlier HCL→public RnD sync; this branch already carries the newer, equivalent content for every conflicting file. I verified each of the 14 conflicting files is byte-identical to the current HCL source-of-truth (`attendants.rs` matches it minus the intended chat-integration strip), while `#864`'s versions are the older snapshot — and no file exists in `#864` that's absent here. Resolved with `-s ours` so the public tree stays exactly the vetted, current state and `#864` is recorded as incorporated. The pushed tree is byte-identical to the previously-validated head, and a secret-scan over the delta is clean.

Issue #868, opened 2026-06-28
[audiopus_sys] build CMake problem and package unmaintained.


AaronNGray:
**Describe the bug**
There is a build problem related to CMake on building [audiopus_sys](https://crates.io/crates/audiopus_sys)

**To Reproduce**
1. Install https://crates.io/crates/videocall-cli

**Unexpected behavior**

```
   Compiling tokio v1.52.3
error: failed to run custom build command for `audiopus_sys v0.2.2`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installcb24AI/release/build/audiopus_sys-8aa97ad7cf534c54/build-script-build` (exit status: 101)
  --- stdout
  cargo:info=No feature or environment variable found, linking by default.
  cargo:rerun-if-env-changed=OPUS_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
[...]

PR #870, opened 2026-06-28
Added Crate.toml patchs for unmaintained audiopus-sys to main project and videocall-cli project

AaronNGray:
Added Crate.toml patchs for unmaintained audiopus-sys to main project and videocall-cli project.

This is a first quick stage in dealing with issue https://github.com/security-union/videocall-rs/issues/868

_Please include a summary of the changes and related context._

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [ ] No tests needed
- [x] `create check`
## Other
- [ ] Documentation updated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,3 +34,6 @@ anyhow = "1.0.75"
 wasm-bindgen-futures = { version = "0.4.30" }
 wasm-bindgen = "0.2.108"
 web-transport-quinn = "0.8.1"
+
+[patch.crates-io]
+audiopus_sys = { git = 'https://github.com/sevenc-nanashi/audiopus_sys', rev = '00e9d16' }
--- a/videocall-cli/Cargo.toml
+++ b/videocall-cli/Cargo.toml
@@ -43,3 +43,6 @@ videocall-nokhwa = { path = "nokhwa", version = "0.10.9", features = ["input-nat
 path = "../videocall-types"
 version = "5.0.0"
 
+[patch.crates-io]
+audiopus_sys = { git = 'https://github.com/sevenc-nanashi/audiopus_sys', rev = '00e9d16' }
+


darioalessandro (commented):
First off — thank you so much for jumping on this! 🙏 You clearly did your homework: tracking down the CMake 4.x policy change, finding the `RUSTSEC-2026-0150` advisory, *and* linking the upstream fix. That's a genuinely high-quality bug report turned into a PR, and we really appreciate you taking the time. ❤️

The diagnosis is spot-on and the fork you're patching to (`sevenc-nanashi/audiopus_sys@00e9d16`) does exactly the right thing — it sets `CMAKE_POLICY_VERSION_MINIMUM=3.5`, which is precisely what the CMake error was asking for. Nice find.

A couple of friendly notes so we land this cleanly:

**1. The `videocall-cli/Cargo.toml` patch can be dropped.** `videocall-cli` is a member of the root workspace, and Cargo only honors `[patch]` in the **workspace-root** manifest — patch sections in member crates are silently ignored ([docs](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html)). The root `Cargo.toml` hunk already covers `videocall-cli`, so the second hunk is a no-op. Keeping just the root one keeps things tidy.

**2. Heads-up on scope (totally fine for a first pass!).** This will fix building the repo from a `git clone` / CI, which is great. The one thing it *won't* reach is the exact repro in the issue — `cargo install videocall-cli` from crates.io — because Cargo strips `[patch]` out of the published manifest, so installers still resolve the unpatched `audiopus_sys 0.2.2`. No action needed from you on that; closing that gap means changing the *published* dependency (a maintained Opus binding) and is a separate piece of work on our side.

For anyone hitting this **right now**, the immediate unblock is to install system libopus — `sudo apt install libopus-dev pkg-config` — since `audiopus_sys` will discover it via pkg-config and skip the vendored CMake build entirely.

If you can drop the `videocall-cli/Cargo.toml` hunk, this is good to go as a build-from-source fix. Thank you again for the careful work and for contributing to videocall-rs — it really does mean a lot! 🚀

darioalessandro:
@AaronNGray I removed that dependency on libopus all together!! check it out https://github.com/security-union/videocall-rs/pull/872

AaronNGray:
@darioalessandro Many Thanks, just finding my feet in rust and its ecosystem, obviously have to look around the area more before proposing a PR.

PR #889, opened 2026-07-17
RnD sync

testradav:
# July 21
### Screen-share reliability overhaul, meeting reactions, client-side recording, and audio-loss observability

**Screen share: the freeze family, fixed end-to-end** (field-diagnosed across two live meetings)
- fix: a static screen share could freeze for receivers and never refresh — the sharer only re-encoded its retained frame when a keyframe request happened to arrive intact, and a receiver whose request was lost held a stale frame forever. The sharer now also emits a retained-frame keyframe on a bounded 3-second wall-clock floor while capture is paused (budget-limited so an idle share costs nothing; the floor state survives encoder restarts), and a receiver stuck in a "deltas flowing, keyframe missing" freeze now escalates on hold-duration instead of a head-age signal that stayed pinned below the ceiling and never fired. Validated by a new e2e spec run green live and proven red on the un-fixed build.
- fix: the relay's keyframe-request rate limiter applied its camera-tuned budget (1/sec per pair) to screen shares — for static content whose only paintable data is keyframes, that throttling produced 13–88 s freezes for every receiver in a field meeting. Screen requests now have their own budget (4/sec steady, 8/sec under congestion) while camera behavior is byte-identical and the global per-receiver cap is unchanged; the sharer-side keyframe coalescer remains the storm backstop.
- fix: a static share oscillated between simulcast layers (~one switch per 2 s — each switch re-arming a keyframe wait and freezing the tile): a congested step-down shortly after an up-switch now latches the layer floor so static content settles on one rung, with the existing time-bounded recovery probing back up; camera layer selection is untouched.
- perf: camera periodic-keyframe intervals are now tier- and transport-aware (relaxed to 7–8 s only on the lowest-bandwidth tiers, hard-capped so receiver-side recovery still bounds the wait), and the static-share keyframe floor fans out to fewer simulcast layers on constrained uplinks — recovery keyframes stop crowding audio on weak connections.

**Audio: WebTransport stall loss made visible and survivable**
- fix: a WebTransport receiver whose main thread stalls could lose whole bursts of audio with no metric anywhere counting it (audio rides unreliable datagrams; the browser's receive queue overflowed invisibly). The client now tracks receive-side audio datagram loss per peer (reorder-tolerant sequence accounting) and reports it through the health pipeline, and the incoming datagram queue is enlarged with a bounded staleness cap so a brief stall delivers audio late-but-complete — mirroring why WebSocket survives the same stall.

**Meeting reactions**
- feat: emoji reactions — a reactions palette in the action bar (keyboard/ARIA toolbar semantics, stays open for rapid repeat reactions, Escape/outside-click/✕ to close), floating reaction overlays on participant tiles with per-sender coalescing and a concurrency cap, screen-reader announcements, and a relay-validated reaction wire type with server-stamped attribution so reactions cannot be forged.

**Client-side recording**
- feat: participants can record a meeting locally to an mp4 saved on their own device (host-only or all-participants per a server toggle; guests never). The recorder composites audio, video tiles, and shared content into a rendered video written in chunks so a crash can't lose the whole recording; everyone in the meeting sees a persistent "Recording" status bar while any recorder is active, with per-session recorder attribution.

**Meeting UX fixes**
- fix: toggling your camera no longer makes your tile jump position in the grid — tiles render from a single join-time-sorted list where camera state decides what a cell shows, not where it sits.
- fix: pinning now works during an active screen share, and a pinned participant no longer gets unpinned when they start talking.
- fix: the self-view's transport badge (WT/WS) no longer overlaps the connection-quality indicator — the self tile's corner chrome (LED, badge, quality pill) is now one flex cluster mirroring the peer-tile convention, and the badge persists through camera on/off.
- fix: on a reconnect, the client re-fetched participant state more than once per real connect during re-election waves; the reseed is now deduplicated per elected connection, trimming redundant load at scale.

**Observability**
- feat: screen-share playout metrics (including content staleness — "how long has the shared content been frozen") now flow end-to-end from the client to per-peer Prometheus gauges with a dedicated dashboard section, so screen freezes can be charted directly instead of reconstructed from logs.

# July 17
### Action-bar overflow, static screen-share delivery, detached-window sizing, and notification-preferences polish

**Action bar: responsive overflow for narrow viewports**
- feat: when the meeting action bar runs out of horizontal room, secondary controls now collapse into an overflow popover behind a trigger button instead of clipping — the core controls (microphone, hang-up) always stay directly visible. Covered by a dedicated e2e spec exercising the overflow behavior across viewport widths.

**Screen share: static content now reaches everyone, immediately**
- fix: a static screen share (a slide, a document, a paused terminal) was invisible to any participant who joined after the content went still — and the detached-content popup started black — until the sharer moved something. Real screen-capture tracks emit frames only on damage, so the encode loop sat parked while keyframe requests flagged a frame that never arrived. The sender now retains the most recent frame and re-encodes it as a keyframe on demand (coalesced to at most one re-encode per 2 s; the GOP stays anchored to real frames), and the detached popup is primed with a no-op repaint at open. Proven live with fails-on-unfixed receipts on both WebSocket and WebTransport.
- feat: the detached shared-content window now opens sized to the content's true aspect ratio (from the decoded resolution) instead of inheriting the in-page pane's box — an ultrawide terminal share opens as a wide window rather than a pane-shaped one with heavy letterboxing. Pure, host-tested sizing math with a per-axis minimum floor; extreme aspect ratios letterbox gracefully via object-fit.

**Devices: better behavior when the camera is blocked**
- fix: further improvements to how the app behaves when media devices are blocked or unavailable, continuing the device-error-handling hardening from the previous cycle.

**Notification preferences: compact announcement matrix**
- feat: the Settings → Preferences → Notifications section is consolidated from four verbose toggle rows into a compact 2×2 matrix — Participant joins / Participant leaves × Message / Sound — with the explanatory text moved into accessible (?) help tooltips (hover, keyboard focus, and screen-reader supported; Escape dismisses the tooltip in place without closing the settings dialog). Each toggle keeps a distinct accessible name, and the underlying settings and their stored values are unchanged.

**Tooling**
- tooling: a new deep meeting-analysis script (`scripts/meeting_quality_xref.py` + README) cross-references meeting console logs through a rule-based anomaly engine, complementing the existing log-parsing front-end.


testradav (commented):
# ✅ RnD sync — July 21 cycle reviewed; merging (squash)

*(Self-PR review posted as COMMENT. Reviewed at head `81ffb5c4`.)*

## 🔍 Sync verification receipts (this cycle)

- **Merge scope**: `labs/PR-staging` → `RnD`; 56 conflicts resolved (11 backend crate files + 31 frontend/docker/e2e files via specialist three-way resolution, each verified as a superset resolution with nothing public-only lost; 14 mechanical). Independent conflict-marker sweep: clean.
- **Chat-strip (HCL-internal Smatter/JMAP)**: reference sweep over `dioxus-ui/src`, `docker`, `e2e` → **zero hits**; 12/12 blocked HCL-only paths verified absent (including the silent-`git rm`-failure class — caught by verification and force-stripped).
- **Workflows**: `.github/workflows/` byte-matched to `upstream/main`'s authoritative set (verified by file-list diff — the 2026-07-16 survivor class can't recur).
- **Builds on the merged tree**: `videocall-ui` wasm ✅, `videocall-client` wasm ✅, `meeting-api` ✅, `actix-api` ✅, workspace `cargo fmt --check` ✅.
- **History-leak gate**: cycle squashed to one commit on the public tip (tree verified byte-identical pre/post squash); secret scan over the full pushed diff: **clean**; upstream-ancestry preserved (`mergeable: true` from creation).
- **Leak gate on the PR file list**: all 105 changed files across every page — zero chat/JMAP/HCL/blocked paths.

## 🤖 CI classification (all four red checks are pre-existing, with evidence)

| Check | This head | Prior head `b064027a` | Classification |
|---|---|---|---|
| ESLint + Prettier + TypeCheck | ✅ success | ❌ failure | **Fixed by this cycle** |
| Docker builds (API, Meeting API, UI, Website) | ✅ ×4 | ✅ ×4 | pass |
| cargo tests_run | ❌ | ❌ | 12 failures, all `webtransport::tests::*` dying on `WebTransport server error: failed to open key file` — the runner lacks TLS fixtures; zero logic/unit failures. Environmental, identical on prior head. |
| WebAssembly Tests | ❌ | ❌ | Workflow passes `--features libvpx,test-utils` to `videocall-codecs`; those features exist only in `upstream/main`'s crate shape, not the fork's. Upstream-vs-fork drift, identical on prior head. |
| Dioxus UI Component Tests | ❌ | ❌ | Workflow expects nix devshell `frontend-tests` absent from the fork's flake. Same drift class, identical on prior head. |
| build_and_package (videocall-cli .deb) | ❌ | (did not run — paths filter) | `env-libvpx-sys` system-probe failure in `videocall-cli` — **zero files of that crate changed in this push** and the dep predates the sync; the job only woke because this cycle touched `videocall-types`/`protobuf`. Pre-existing crate/runner drift. |

Net: this cycle *improves* the PR's CI (one check red→green, zero new failures) and every remaining red pre-dates it. Prior sync cycles merged with the same drift set.

**Verdict: approve — merging with squash per upstream convention.**

PR #890, opened 2026-07-20
macOS camera integration rewrite: Swift AVFoundation capture layer replacing objc nokhwa bindings

darioalessandro:
## Summary

Replaces the 2,417-line hand-written `objc` `msg_send!` AVFoundation binding in the vendored nokhwa fork (`videocall-nokhwa-bindings-macos`) with a real Swift capture layer bridged over a hand-written C ABI. The `videocall-nokhwa` public API is unchanged; `videocall-cli` needed only graceful-error handling changes.

The old binding's own author left this comment in it: *"whatever is written here will induce horrors uncomprehendable... write apple code in swift and bind it to rust."* This PR does exactly that.

### Architecture

```
videocall-cli (unchanged API surface)
  └─ videocall-nokhwa facade (Camera / query / Buffer<I420>)
      └─ avfoundation.rs (CaptureBackendTrait, blocking frame() via bounded flume channel)
          └─ src/apple.rs — safe RAII wrappers over extern "C"          [NEW]
              └─ 8-function @_cdecl `vcc_*` C ABI                      [NEW]
                  └─ SwiftPM static lib `VideocallCapture`             [NEW]
                     (macOS 12+ / iOS 15+, built per-target by build.rs)
```

- **No third-party interop crates** — hand-written `@_cdecl` ↔ `extern "C"` only (plain C ABI, the boundary both Apple and Rust guarantee).
- Bindings crate deps reduced from 7 native crates (`objc`, `cocoa-foundation`, `block`, `core-media-sys`, `core-video-sys`, `core-foundation`, `once_cell`) to `nokhwa-core` + `flume`.
- `build.rs` cross-builds the Swift package for macOS (arm64/x86_64), iOS device, iOS simulator (incl. Intel), and Catalyst; no-op on non-Apple targets.
- Swift patterns: serial session queue with `DispatchSpecificKey` guard, `beginConfiguration`/`commitConfiguration` bracketing, exact-resolution-or-error format selection, CMTime-clamped frame-rate resolution, tight plane compaction (`FramePacker`).

### Bug fixes over the old backend

- NV12 was mapped to `kCVPixelFormatType_420YpCbCr10BiPlanarVideoRange` (**10-bit**); now correct 8-bit `'420v'`/`'420f'`.
- Every frame was tagged `FrameFormat::GRAY`; now tagged with the real per-frame format from the delivered buffer's OSType.
- 420-biplanar fourccs were misclassified as YUYV.
- `videoDataOutput.videoSettings` now pins pixel format **and** width/height — AVFoundation otherwise delivers session-preset geometry instead of `activeFormat` geometry, which corrupted frames (chroma banding + doubled image).
- `nv12_to_i420`/YUYV converters return errors instead of panicking on size/format mismatches.
- Unbounded frame channel → `bounded(2)` drop-on-full (hard memory cap under encoder stalls); one full-frame copy removed via zero-copy `Buffer::new_from_vec`.
- Camera setup failures print a clean error and exit(1) instead of panicking the capture thread; hot-loop decode failures log-and-skip with exit after 30 consecutive.

### Tests

- **Swift (16)**: frame-rate resolution (incl. the UVC 59.99976 case), fourcc mapping, padded-stride packing oracle with 0xEE padding sentinel.
- **Rust (18)**: FFI fourcc mapping, per-range fps enumeration regression, bounded-channel drop semantics, decode guards (short/mis-tagged/odd-dim buffers → Err, never panic).

### Verification

- `swift test` 16/16, `cargo test` green across nokhwa-core/bindings/videocall-cli, `cargo clippy --all-targets` clean, `cargo fmt` clean.
- iOS: `swift build` for arm64-apple-ios + `cargo check --target aarch64-apple-ios` green (capture layer compiles for iOS; no iOS consumer wired yet — building toward native).
- Live: MacBook Pro Camera @720p and OBS Virtual Camera @1080p stream visually clean (pre-encode I420 frames dumped and inspected); unsupported resolutions fail with a descriptive error pointing at `info --list-formats`.


darioalessandro:
## 1/7 — What was there before (and why it had to go)

The macOS camera code lived in one file: `nokhwa-bindings-macos/src/lib.rs` — **2,417 lines of Rust pretending to be Objective-C**. It used the `objc` crate's `msg_send!` macro to send raw ObjC messages to AVFoundation, one unchecked call at a time:

```rust
let session: *mut Object = unsafe { msg_send![class!(AVCaptureSession), new] };
```

Every call was `unsafe`, nothing was type-checked, and the whole file sat under `#![allow(warnings)]`. The original author knew it was bad — their own comment at the top said: *"whatever is written here will induce horrors uncomprehendable... write apple code in swift and bind it to rust."*

It also had real bugs baked in:
- NV12 was mapped to a **10-bit** pixel format constant (should be 8-bit `'420v'`)
- Every frame was tagged `GRAY` no matter what it actually was (with a `// FIXME: dont, what the fuck???`)
- The frame channel's sender was smuggled through an ObjC instance variable as a raw pointer
- FPS selection was documented as "does not work"

This PR takes the author's advice literally: the Apple code is now written in Swift, and Rust binds to it.

darioalessandro:
## 2/7 — The new shape: three clean layers

The replacement splits the one giant unsafe file into three layers, each doing one job:

```mermaid
graph TD
    CLI["videocall-cli<br/><i>unchanged — same Camera / query / frame() API</i>"]
    RS["<b>Layer 3 — Rust</b>: src/apple.rs (~500 lines)<br/>safe RAII wrappers: open on new, close on Drop"]
    ABI["<b>Layer 2 — C ABI</b>: 8 functions, vcc_* prefix<br/>the only place Swift and Rust touch"]
    SW["<b>Layer 1 — Swift</b>: swift/Sources/VideocallCapture/<br/>real AVFoundation code: CaptureEngine, DeviceDiscovery,<br/>PixelFormatMapping, FrameRateResolver, FramePacker"]
    AVF["Apple AVFoundation / CoreMedia / CoreVideo"]
    CLI --> RS --> ABI --> SW --> AVF
```

**Layer 1 (Swift)** owns everything Apple-specific: the capture session, device discovery, format selection, permissions, and packing pixel buffers into tight byte arrays. It's a normal SwiftPM package — you can open it in Xcode, run `swift test` on it, and it builds for macOS 12+ **and** iOS 15+.

**Layer 2 (the C ABI)** is the deliberately tiny waist: `vcc_authorization_status`, `vcc_request_access`, `vcc_enumerate_devices`, `vcc_enumerate_formats`, `vcc_capture_open`, `vcc_capture_start`, `vcc_capture_stop`, `vcc_capture_close`. Only C-representable things cross it: integers, pointers, byte buffers. No third-party interop crate — plain `@_cdecl` on the Swift side, plain `extern "C"` on the Rust side, the one boundary both Apple and Rust guarantee stability for.

**Layer 3 (Rust)** wraps those 8 functions in safe types: `CaptureStream` opens the device in its constructor and guarantees stop/close/free happens exactly once in `Drop`. No `unsafe` leaks above this layer.

The old `lib.rs` shrank from 2,417 lines to 37 — it's now just a platform gate that re-exports `apple.rs`.

darioalessandro:
## 3/7 — How a camera opens now

When `videocall-cli` asks for "camera X at 1280×720 NV12 @ 30fps", here's the sequence:

```mermaid
sequenceDiagram
    participant CLI as videocall-cli
    participant RS as Rust (apple.rs)
    participant SW as Swift CaptureEngine
    participant AV as AVFoundation

    CLI->>RS: Camera::new + open_stream()
    RS->>SW: vcc_capture_open(uniqueID, 1280, 720, NV12, 30)
    SW->>AV: find device by uniqueID
    SW->>SW: selectFormat: EXACT 1280x720 match in device.formats<br/>— no match → return null → clean Rust error
    SW->>AV: beginConfiguration
    SW->>AV: activeFormat = matched format
    SW->>AV: videoSettings = { NV12, width: 1280, height: 720 }
    SW->>SW: FrameRateResolver: nearest supported fps range,<br/>clamped with the range's own CMTimes
    SW->>AV: commitConfiguration → startRunning
    SW-->>RS: opaque handle
    RS-->>CLI: Ok — frames about to flow
```

Three decisions worth knowing:

1. **Exact resolution or clean error.** The old code (and an early version of this PR) would silently fall back to a "close enough" resolution — which crashed the decoder downstream when the bytes didn't match expectations. Now: if the device doesn't offer exactly what you asked, `open` fails with an error telling you to run `info --list-formats`. Same contract the old backend intended, but enforced where it's reliable — against the device's real format list.

2. **`videoSettings` pins format AND dimensions.** This is the subtle one: AVFoundation can deliver buffers at the *session preset's* geometry rather than the `activeFormat` you selected. Pinning width/height in `videoSettings` forces delivery at exactly the negotiated size. (Comment 6 shows what happened without this.)

3. **Frame rates are clamped, not compared.** External/UVC cameras report ranges like "59.99976 fps". Asking for integer 60 used to throw an ObjC exception. The resolver picks the nearest supported range and clamps the request into it using the range's own `CMTime` values — a pure function with its own unit tests.

darioalessandro:
## 4/7 — The journey of one frame

Once the session runs, every frame travels this path (30–60 times per second):

```mermaid
sequenceDiagram
    participant AV as AVFoundation<br/>(camera hardware)
    participant SW as Swift<br/>captureOutput
    participant CB as C callback
    participant RS as Rust<br/>frame_trampoline
    participant CLI as videocall-cli<br/>camera thread

    AV->>SW: CMSampleBuffer (CVPixelBuffer inside)
    SW->>SW: lock buffer (read-only)
    SW->>SW: FramePacker: copy planes into reused scratch,<br/>stripping row-padding (bytesPerRow → tight bytes)
    SW->>SW: unlock buffer (returned to AVFoundation's pool ASAP)
    SW->>CB: frame_cb(ctx, bytes, len, w, h, REAL fourcc)
    CB->>RS: extern "C" → Vec<u8>
    RS->>RS: try_send into bounded(2) channel<br/>(full? drop frame — encoder is behind anyway)
    CLI->>RS: frame() — blocking recv()
    CLI->>CLI: NV12 → I420 conversion (reused buffer)
    CLI->>CLI: VP9 encode → WebTransport → meeting
```

Details that matter:

- **Tight packing.** Cameras pad each pixel row to alignment boundaries (`bytesPerRow` > useful bytes). `FramePacker` strips that padding so Rust always receives exactly `width × height × bpp` bytes — and it's a pure enum with oracle tests that plant a `0xEE` sentinel in the padding and assert it never leaks into output.
- **The format tag travels WITH the frame.** Each callback carries the fourcc read from that specific buffer — not the format we *requested*. The old code tagged everything `GRAY` and hoped. Now bytes and label can't disagree.
- **Bounded channel = bounded memory.** The old channel was unbounded: if the VP9 encoder stalled for one second, ~80 MB of frames piled up. Now capacity is 2 and the capture side drops-on-full — correct, because the consumer only ever wants the freshest frame.
- **Teardown is race-free.** `stop()` clears the delegate on the session queue, runs a sync barrier on the data queue to drain any in-flight callback, and only then does Rust free the callback context. Verified in review: no use-after-free window, context freed exactly once on every path.

darioalessandro:
## 5/7 — How the build works (Rust compiles Swift now)

`cargo build` alone produces the whole thing — no Xcode project, no manual steps:

1. `nokhwa-bindings-macos/build.rs` detects an Apple target and runs `swift build -c release` on the SwiftPM package, choosing the right triple + SDK for whatever cargo is building:
   - `aarch64-apple-darwin` / `x86_64-apple-darwin` → macOS SDK
   - `aarch64-apple-ios` → iPhone SDK
   - `aarch64-apple-ios-sim` **and** `x86_64-apple-ios` (Intel sim has no `-sim` suffix — classified by arch) → simulator SDK
   - `*-macabi` → Catalyst
2. It links the resulting `libVideocallCapture.a` static library plus the AVFoundation/CoreMedia/CoreVideo frameworks and the Swift runtime search paths into the Rust binary.
3. Swift is **always** built `-c release` (whole-module optimization) — even in debug cargo builds, the frame hot path stays fast.
4. On Linux/Windows the build script is a no-op — CI for other platforms is untouched.

Build output goes under cargo's `OUT_DIR` (per-target isolation), and `swift/.build/` is gitignored, so dev `swift test` runs never pollute the repo.

Why this matters beyond macOS: the same package already compiles for iOS (`cargo check --target aarch64-apple-ios` passes). Nothing on iOS consumes it yet, but when the native iOS app wants real camera capture through this same Rust pipeline, the capture layer is already there — same pattern as the existing VideocallCodecs/VideoCallKit xcframeworks.

darioalessandro:
## 6/7 — The two bugs live testing caught (and what they taught us)

Unit tests all passed — then real cameras found two bugs. Both are worth understanding because they're classic capture-pipeline failure modes.

**Bug A: the `mid > len` panic.** Requesting 1280×720 from OBS Virtual Camera (which only outputs 1920×1080) crashed with `panicked at types.rs:1769: mid > len`. Chain of events: an early "nearest-larger" fallback silently opened the camera at 1080p → frames arrived carrying ~3.1 MB → but the consumer had already sized its I420 buffer for 720p (~1.4 MB) from the *requested* resolution → the plane-split ran off the end. Two-part fix: the exact-resolution-or-error contract (comment 3), and the NV12→I420 converter now returns a proper `Err` on any size mismatch instead of panicking — with regression tests for short-input, short-output, mis-tagged-format, and odd-dimension cases. A frame can never crash the decoder again.

**Bug B: corrupted pixels.** Camera opened, stream flowed, but the picture was magenta/green banding with the scene doubled side-by-side. Root cause: AVFoundation delivered buffers at the **session preset's geometry** instead of the selected `activeFormat` — the decoder read rows at ~2× the expected width. The debugging method is the reusable lesson: dump one raw frame at each pipeline stage, brute-force decode it under *every* candidate interpretation (`nv12`, `yuv420p`, `yuyv422`, `uyvy422`, `bgra` — one ffmpeg command each), and look at the PNGs. The interpretation that renders clean IS the true encoding; comparing it against each stage's assumption pinpoints exactly where the lie enters. Fix: pin width/height in `videoSettings` (comment 3, decision 2).

Why tests missed both: every real Mac camera tested delivers **contiguous, unpadded** planes, so padded-stride and geometry-mismatch code paths never executed. That gap is now closed with synthetic oracle tests that fabricate padded/mismatched buffers on purpose.

darioalessandro:
## 7/7 — What stayed the same, what got verified, what's next

**Unchanged:** the `videocall-nokhwa` public API — `Camera::new`, `query()`, `nokhwa_initialize`, `RequestedFormat`, blocking `frame()`, `decode_image_to_buffer::<I420Format>`. Cameras are still addressed by their uniqueID UUID on macOS. The Windows and Linux backends are untouched. `videocall-cli`'s only changes are user-facing polish: camera setup failures now print a clean one-line error and exit(1) instead of a thread panic with a backtrace, and a transient decode failure logs-and-skips instead of killing a live call (clean exit only after 30 consecutive failures).

**Verified:**
| Check | Result |
|---|---|
| `swift test` (frame-rate resolver, fourcc mapping, FramePacker oracles) | 16/16 |
| `cargo test` (nokhwa-core 7, bindings 4, videocall-cli 7) | 18/18 |
| `cargo clippy --all-targets` / `cargo fmt` | clean |
| `swift build` arm64-apple-ios + `cargo check --target aarch64-apple-ios` | green |
| Live: MacBook Pro Camera @720p, OBS @1080p | streams visually clean |
| Live: unsupported resolution (OBS @720p) | clean error + exit 1, no panic |

Dependency diet: the bindings crate went from 7 native crates (`objc`, `cocoa-foundation`, `block`, `core-media-sys`, `core-video-sys`, `core-foundation`, `once_cell`) to **2** (`nokhwa-core`, `flume`). Net diff: +2,067 / −2,724 — smaller than what it replaced, tests included.

**Known follow-ups (deliberately out of scope):**
1. The frame-pacing loop in `producers/camera.rs` busy-spins between frames (pre-existing; a one-line `thread::sleep` would save battery)
2. `info --list-formats` groups some formats under the wrong header (pre-existing `!=` filter, one-character fix)
3. Wire the (already iOS-ready) capture layer into the native iOS app

PR #895, opened 2026-08-04
RnD sync — August 4

testradav:
# August 14
### Meeting-log tooling that reports what it can actually support, and a cluster-agnostic postgres chart

- tools: error census in the meeting-log parse tool — error signatures are now grouped and ranked by how many **distinct participants** hit them. The Sessions table's `Errors` column is a count, so a defect several people share just inflates one row and reads as that one person's problem; distinct-participant ranking is what separates a widespread defect from a noisy session.
- tools: the `Speak` column is open-mic energy crossing a threshold, **not** evidence of who was talking. Three places in the parse tool and its README described it as the latter, which it cannot support; the descriptions now say what the signal actually is.
- test: a 26-case suite for the meeting-log parse tooling, plus a `make test-scripts` target to run it — the bash/python tooling under `scripts/` had no automated coverage at all, so a regression there would have rotted silently.
- helm: a cluster-agnostic `videocall-postgres` chart — a thin wrapper pinning `bitnami/postgresql` that carries only the settings every deployment shares. Its values schema deliberately fails fast (`lint`, `template`, `install` and `upgrade` all error) unless an environment values file supplies the database name, storage class, size and resources, so no cluster-specific assumption can hide in the chart. Consolidates per-cluster wrapper charts that had drifted apart.
- docs: the review rules now gate comment **volume**, not just comment accuracy — the acceptance criteria and the adversarial self-review rule both constrained what a comment may *claim* while saying nothing about how many claims a diff may add.

# August 13
### Speaker-glow correctness, sharp screen-share text, and honest receiver-side measurement

- fix: the highlight-speaker glow no longer sticks on a silent peer — four distinct defects were stacked behind the one symptom: a diagnostics subscriber that died permanently on bus overflow, a receiver that ignored the sender's own VAD flag, a mute path that never emitted a terminal zero, and an unbounded tail now closed by a deadman timer
- fix: close the mute race at the producer rather than the consumer, veto straggler level events that arrive after a mute, and stop the decoder-reset blink
- fix: give the glow a single definition of "dark", tied to the write gate that decides whether a level can be maintained at all — previously the resolver and the deadman disagreed, stranding a tile lit at the bottom of the ramp
- perf: rebuild only the decoder that actually failed instead of all three — a receiver holding a stale key was respawning the NetEQ audio worker tens of times a second, on errors raised before any decoder ran; also closes two closure leaks and replaces per-sample wasm↔JS boundary crossings in the RMS path with one bulk copy
- fix: sharp screen-share text — capture 1:1 at native HiDPI resolution and choose the starting quality tier from the source, instead of destroying three quarters of the pixels through two stacked fractional downscales
- fix: receive-side fps counts DECODED frames rather than arrivals — it had been reading the simulcast ladder sum, so a low-fps multi-rung camera reported several times its real rate
- fix: the peer tile shows the audio rung actually being decoded instead of the ladder base, so a receiver on a higher rung no longer reads 12k when it is decoding 24k or 48k
- feat: record keyframe ARRIVAL at the receiver — the fact that distinguishes "never delivered" from "arrived but didn't recover"
- fix: publish the fitted encode geometry so the self-view stops reporting the quality-tier box instead of the real encoded frame
- feat: export fitted per-layer camera encode geometry and measured per-layer output fps, with dashboard panels including real throughput
- feat: typed custom meeting-timer duration — an arbitrary value in seconds or minutes, validated before it reaches the wire, alongside the existing one-click presets
- fix: un-clip the reactions-palette close badge, which a later overflow rule had begun clipping, and remove the spurious horizontal scrollbar
- a11y: the pinned SEND quality floor announces the base rung rather than the top of the ladder
- fix: guard gap-detection against non-monotonic input, with high-RTT end-to-end coverage
- fix: bound the bot dashboard's control proxy with an SSE-safe inactivity timeout, so a silent handler can no longer hang the request and the operator's tab forever
- test: bot video sources publish 640x480 square-pixel — the geometry the overwhelming majority of real publishers actually send; bots had been generating roughly three times a real user's pixel load on encode and on every receiver
- test: bots count the video rung they would decode rather than every rung the relay forwards, render at an explicit 1080p desktop, and advertise the shipped three-rung simulcast ladder
- fix: stop tracking the ephemeral WebTransport dev-certificate hash, which every local end-to-end run rewrote, and fail fast when it is absent
- test: serialize the process-global counters that host tests inject into, and correct doc comments that contradicted the code

# August 4
### Meeting password, host timer, raise hand, identity hardening, and bot-fleet/CI reliability

- feat: meeting passwords — enforced server-side, with a password prompt in the join flow
- feat: host-controlled meeting timer — countdown visible to the whole room; the host can start, extend, or cancel it
- feat: raise and lower hand, with a persistent banner so a raised hand stays visible even when the participant has no video tile
- feat: search field in the emoji reactions picker, ranked and capped
- feat: runtime-selectable reduced camera ladder (180p/360p/540p) plus publisher-side screen-encoder stall observability
- fix: stamp authenticated identity onto the broadcast packet wrapper server-side; fail closed on unparseable frames
- fix: server-stamp health-packet identity, clamp gauge ingest to finite values, and bound client-supplied metric label cardinality
- fix: let the video-fps metric report a genuine zero instead of latching its last healthy value
- fix: re-anchor the audio datagram-loss tracker when the audio layer switches
- fix: thread the selected ladder variant into receiver-side quality rung labels
- fix: filter self packets against the full session-id history so stale-session echoes are dropped
- fix: cross-window rate limiting and instance-level dedupe for presence re-announce storms
- fix: stop attendants listener/timer leaks and make peer-tile memoization actually engage
- fix: bind the viewport resize listener to the component lifetime (listener leak on unmount)
- a11y: the dock menu is now a named command menu with the standard menu keyboard model; closed several customize-mode accessibility gaps
- a11y: the Decay help affordance is keyboard-operable, readable in both themes, and unclipped
- fix: dock popover surface is genuinely fixed-dark in both themes
- tests: un-skipped the client wasm suite in CI, repaired 23 rotted tests, and fixed the backwards-clock wedge they exposed
- tests: locked API/UI contracts — unauth-401 on the feed and joined routes (with authenticated-200 controls), sign-in visibility for anonymous profiles, device-refresh callback safety, real heartbeat tick observation
- bots: K8s bot fleet Increment 2 — StatefulSet with per-ordinal identity and capability spoofing; per-bot PVC for run artifacts; bounded SSH host probe and remote CSV retrieval; control token moved off the retained volume; benign teardown race no longer logged at error level
- ci: shared-runner hardening — scoped process cleanup so jobs can't kill siblings, and fixed a leaked child process holding the wasm test step's stdout open
- docs: corrected shipped comments that contradicted the code (audio simulcast default, relay layer filtering) and recorded the encoder-stall observability gap


testradav:
Pushed the **August 13** sync (`908a7168`) — a new dated section is at the top of the PR body.

Headline items: the speaker-glow stuck-highlight chain is fully closed (four stacked defects, plus three follow-ups found by an adversarial re-review of the merged fix), screen-share text is sharp again (1:1 HiDPI capture instead of two stacked fractional downscales), and several receiver-side measurements now report what they claim to — decoded-frame fps rather than the simulcast ladder sum, and the audio rung actually being decoded rather than the ladder base.

testradav:
Pushed the **August 14** sync (`fde6b7b4`) — new dated section at the top of the PR body.

Smaller cycle than the last one, and it's all tooling/docs/infra rather than client changes: the meeting-log parse tool now ranks error signatures by distinct participants (a count column made a shared defect read as one person's problem) and no longer claims its `Speak` column shows who was talking — it's open-mic energy crossing a threshold. Also adds the first automated coverage for the `scripts/` tooling (26 cases behind `make test-scripts`) and a cluster-agnostic `videocall-postgres` helm chart whose values schema fails fast rather than letting a cluster-specific default hide in it.

PR #902, opened 2026-08-11
Vlog: concrete conclusions — you can uninstall Docker Desktop

darioalessandro:
The post's lessons were tangential aphorisms. Replaced with what matters to a reader: no Docker Desktop needed, `make dev` installs nix and runs everything natively, Docker only remains for prod-image runs and e2e; links the migration guide.


darioalessandro:
Folded into the consolidated PR

PR #909, opened 2026-08-13
docs: fix broken star history chart

Dessalines39394:
The star history chart in the README is currently broken and no longer renders due to GitHub stargazer API restrictions. This change repoints it to star-history.dera.page, whose data source requires no API token, restoring the chart for visitors.


darioalessandro:
Thanks for your contribution!

Issue #756, opened 2026-03-12
Optimize performance capabilities


testradav:
videocall needs to perform in deployments where client performance may be degraded due to communication infrastructure not being optimal.  Videocall should adapt and step down audio and video in order to squeeze the maximum out of the conditions. It should also step up automatically when communication conditions improve.  Should mitigate when a peer may have connection on and off (eg. auto join the peer if they go off line momentarily)

PR #913, opened 2026-09-10
fix(neteq): treat DTX silence gaps as silence, not packet loss

mowtee:
When a DTX-enabled encoder (e.g. Opus with DTX, as proposed for the send side in #756) goes silent, it deliberately stops sending frames: sequence numbers stay contiguous while timestamps jump, with only a sparse comfort-noise refresh (~1 per 400 ms). NetEq's decision logic has no notion of this, so a silence gap is handled as packet loss: expand (concealment) runs for the entire gap, `consecutive_expands` ratchets into the ~6 s safety-valve `reset()` on every longer pause, and refresh inter-arrival times feed the delay manager as if they were network jitter. With #756's sender-side DTX in place, every receiver will hit this on every silence.

**Changes**

- `insert_packet()`: detect DTX (contiguous sequence + timestamp jump beyond the packet's declared duration); normal cadence ends the gap. Refresh packets no longer update the delay manager — their 400 ms spacing says nothing about network jitter.
- `insert_packet()`: a set RTP **marker bit** (RFC 3550/3551: first packet of a talkspurt) ends a recognized gap immediately — an explicit resume signal for senders that provide it; the cadence rules remain the fallback for those that don't (including this project's protobuf transport, which has no marker field).
- `get_decision()`: during a recognized gap, decode buffered refreshes normally and return `Operation::ComfortNoise` when the buffer is starved — activating the existing (previously unreachable) `generate_comfort_noise()` path. This also breaks an expand trap: with `consecutive_expands > 0` and a below-low-limit buffer, the old logic kept returning `Expand` even when a decodable packet was waiting.
- `statistics`: `operation_counts` was `[u32; 9]` for 12 `Operation` variants — recording `ComfortNoise`/`Dtmf`/`Undefined` panicked out of bounds (latent while those were unreachable). Sized to 12, and the per-second rate mapping now uses `record_decode_operation`'s indices; it previously read a different layout, misattributing e.g. `comfort_noise_per_sec` to `FastAccelerate` counts.

The statistics changes are bundled because the ComfortNoise path panics without the array fix; happy to split the index-mapping correction into its own PR if you'd prefer.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement
- [ ] Breaking change
- [ ] Project / Infra

## Testing
- [ ] No tests needed

Four new tests: a DTX gap plays comfort noise and never runs concealment; a marker-bit resume decodes immediately; a sequence gap (real loss) still expands; 150 DTX refreshes leave the delay target untouched. Full `cargo test -p neteq`: 72 passed. `cargo fmt` + `cargo clippy -D warnings` clean.

## Other
- [ ] Documentation updated

(Behavior documented in code comments.)

Context: found while integrating this crate as the adaptive playout for a native Rust WebRTC client that receives real Opus DTX streams (RFC 7587 transmit semantics).


darioalessandro:
Hey chief!! Thank you for your contribution. I'll take a look asap

